PR #132, opened 2026-07-24
feat(mapping): fs displacement discipline — brand macro-consensus + cross-source save margin

evawlve:
## What

Part 2 of the fatsecret hardening, stacked on #131 (parity). Part 1 lets fatsecret compete fairly; this PR makes sure the extra reach cannot re-open the plausible-wrong-data path that regressed n-mq-08 (a fatsecret Quest bar with "42% protein" that passes every absolute plausibility floor — fiber-bar calorie discounts are legitimate, so no floor can catch it).

1. **Rank-time: brand macro-consensus outlier demotion** (`simple-rerank.ts`). Under the decisive-brand gate, with ≥3 same-brand nutrition-bearing siblings that cover a non-brand query token, if a real majority agrees within 15% of the median panel, a sibling deviating >20% on protein/100g or >30% on kcal/100g gets −0.20. Source-agnostic — it demotes a corrupt OFF per-serving-panel row exactly like a bad fs record. Kill-switch: `RANK_BRAND_CONSENSUS=0` (documented in `.env.example`).
2. **Save-time: cross-source displacement margin** (`validated-mapping-helpers.ts`). An `off:` ↔ `fs:` target change on a non-corrupt incumbent requires challenger confidence ≥ incumbent `aiConfidence` + 0.05. An already-good OFF pick keeps its cache key unless fatsecret is meaningfully more confident — and vice versa (fs incumbents are protected from OFF churn symmetrically). Same-family swaps keep today's supersede-stale semantics; corrupt incumbents forfeit, mirroring the serving-downgrade guard.

## Testing

- New `brand-consensus-outlier.test.ts` (6 cases): quest-class fs outlier flipped, kill-switch control proves the penalty is what flips it, OFF-outlier symmetry, <3-sibling no-op, no-decisive-context no-op, disagreeing-pack no-op.
- `validated-mapping-save-gates.test.ts` +6 margin cases incl. fs-incumbent symmetry, corrupt forfeit, legacy null-confidence rows.
- Full jest: 1232 passed; only failures are the 2 pre-existing live-DB-dependent tests (present on master).
- `lint:ci` 0 errors, `typecheck` clean, `next build` green, env-example parity check passes.
- Box eval gate (flag-off parity, then flag-on expecting n-mq-08 recovered + n-mq-10 kept) before merge; results will be posted here.

https://claude.ai/code/session_01Qr5Tr5RTDXeBTfkX5kE67y

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -220,6 +220,13 @@ OFF_BARE_SERVING_GUARD=1
 # Kill-switch: "0" restores prior ordering.
 RANK_PLAUSIBILITY_PARTITION=1
 
+# Brand macro-consensus outlier demotion (fs displacement hardening, Jul 2026):
+# under the decisive-brand gate, when >=3 same-brand nutrition-bearing siblings
+# mostly agree on the per-100g panel, the rerank demotes a sibling whose
+# protein/kcal deviates hard from the median (plausible-wrong label data).
+# Kill-switch: "0" disables the pass.
+RANK_BRAND_CONSENSUS=1
+
 # Human-triage cache-row trust (PR D pt3, Lever B): FoodMapping rows with
 # validatedBy='human-triage' skip name-heuristic cache escapes at read time
 # (nutrition-invalid and serving-shape escapes stay active). Kill-switch: "0".
--- a/src/lib/mapping/__tests__/brand-consensus-outlier.test.ts
+++ b/src/lib/mapping/__tests__/brand-consensus-outlier.test.ts
@@ -31,6 +31,23 @@ function rerank(candidates: RerankCandidate[]) {
 }
 
 describe('brand macro-consensus outlier demotion', () => {
+    it('fires on the exact n-mq-08 phrasing, where the flavor token breaks decisive adjacency', () => {
+        // "1 quest chocolate chip protein bar": "quest" neighbors "chocolate",
+        // so hasDecisiveBrandContext is false — the pass must not depend on it.
+        const candidates = [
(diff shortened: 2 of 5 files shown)


evawlve:
## Box eval gate results (OptiPlex, branch `7748694`)

**Flag-off** (`FATSECRET_RETRIEVAL_ENABLED` unset): ❌ 1 real failure = exactly **n-mq-10** — identical to the historical baseline. Bonus: known issue **n-serv-57** (kirkland protein bar) now passes via the consensus pass.

**Flag-on**: ✅ **0 real failures** — first fully-green flag-on eval ever.
- **n-mq-08 recovered**: maps to Quest OFF record "Oatmeal Chocolate Chip Protein Bar", 60g label serving, conf 0.95 (the 42%-protein fs record is consensus-demoted). Note: flavor is "oatmeal chocolate chip" rather than plain chocolate chip — in-band on every assertion, but a flavor-precision nuance to keep an eye on.
- **n-mq-10 kept**: fs "Protein Cookie", 60g, conf 0.93.
- Known issues now passing flag-on: **s-supp-11** (ryse cinnamon toast crunch), **n-serv-57**, **n-mq-22** (chicken breast retrieval gap).

The gate caught two calibration bugs along the way (both fixed in this branch): the pass was gated on decisive-brand adjacency which the exact n-mq-08 phrasing defeats (`903ff3d`), and the 0.20 penalty lost to the max ORIGINAL_SCORE spread + fs prior by 0.01 (`7748694`).

Box restored to master + flag-off after the gate.

PR #143, opened 2026-07-25
fix(save-gate): waive the cross-source margin for unbranded whole foods

evawlve:
Funnel fix 3. The cold warm blocked 19 saves on `cross_source_margin`. **All 19 were fatsecret**, and 15 should have cached:

| query | picked | kcal/100g |
|---|---|---|
| grilled tilapia | Tilapia (Cooked, Dry Heat) | 128 |
| boiled egg | Boiled Egg | 154 |
| brussels sprouts | Brussels Sprouts | 43 |
| baked sweet potato | Sweet Potato (Baked In Skin) | 90 |
| steamed broccoli | Cooked Broccoli (Fat Not Added) | 35 |
| grilled salmon | Cooked Salmon | 139 |
| sauteed spinach | Cooked Spinach (from Fresh) | 40 |

Several are near-exact USDA values.

## Why it fires

The margin was added as conservatism **before** we had data on how the new source performs. We have the data now. The casualty class is exactly where fatsecret is strongest and OFF is weakest: an **unbranded whole-food query**, where OFF can only offer a packaged product that fits the query worse — yet holds the cache row because it got there first.

## The waiver is narrow

It requires **all** of: the query names no brand, the challenger carries no brand, and the challenger's macros are non-degenerate. A branded query is exactly where the margin still earns its keep, so it keeps it.

## The 4 correct blocks were a different signal

They weren't a source problem at all — they were **preparation mismatches**: `baked chicken tenders` → a *Fried* record, `poached egg` → *Fried Egg*. That's orthogonal to which corpus a record came from, so it becomes its own predicate rather than riding on the margin.

It only fires when the query is **explicit** about preparation (a bare "chicken tenders" is a request for tenders, not a mismatch), and a record that repeats the requested preparation clears it — "Roasted Broiled or Baked Chicken Thigh" is a fine answer to "grilled chicken thighs".

## Known gap, left alone deliberately

`steamed carrots` → `Cooked Carrots` (54 kcal, 2.48g fat) now waives. The triage flagged it, but **the same record is correct for `roasted carrots`** — so separating them needs a fat check against the produce baseline, not a name check. Not worth the regression risk in this PR.

## Verification

- 5 new tests; save-gates suite 52/52; full suite 1065 passing, same single pre-existing failure as master.
- `typecheck` clean, `lint:ci` 0 errors.
- Box eval gate to follow before merge.

### DIFF
--- a/src/lib/mapping/__tests__/validated-mapping-save-gates.test.ts
+++ b/src/lib/mapping/__tests__/validated-mapping-save-gates.test.ts
@@ -651,14 +651,13 @@ describe('cross-source margin waiver for unbranded whole foods', () => {
         expect(mockFoodMappingUpsert).toHaveBeenCalledTimes(1);
     });
 
-    it('waives for "brussels sprouts", whose bare food word collides with a grocery chain', async () => {
+    it('waives for a prep-qualified query whose food word collides with a grocery chain', async () => {
         // detectBrandInQuery reports matchedBrand "sprouts" (Sprouts Farmers
         // Market) here, so a bare detectedBrand check would veto the waiver on
-        // an unmistakable whole food. Production incumbent was an OFF
-        // store-brand row ("Freedom's Choice") holding the key.
-        mockFoodMappingFindUnique.mockResolvedValue({ ...incumbent, normalizedForm: 'brussel sprout' });
+        // an unmistakable whole food. hasDecisiveBrandContext is what keeps it.
+        mockFoodMappingFindUnique.mockResolvedValue({ ...incumbent, normalizedForm: 'roasted brussel sprout' });
         await saveValidatedMapping(
-            'brussels sprouts',
+            'roasted brussels sprouts',
             makeMapping({ foodId: 'fs_5', foodName: 'Brussels Sprouts', brandName: undefined }),
             { confidence: 0.93 } as AIValidationResult,
             { nutrientsPer100g: { kcal: 43, protein: 3.4, carbs: 9, fat: 0.3 } },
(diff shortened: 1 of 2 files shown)


evawlve:
## Closing: the waiver cannot be made safe at this layer

Two box gate runs, and the blocker turns out to be cache-key granularity, not gate tuning.

**Attempt 1 — waive for any unbranded whole food.** Repointed the eight most-used generic
keys in the cache (`egg` 926 uses, `broccoli` 530, `salmon` 342, `chicken` 232, plus
`kale`/`pretzel`/`tilapia`/`brussel sprout`) and broke five golden cases: `two eggs` billed
84g against an expected [88,140], because FatSecret's *Boiled Egg* is a 42g medium egg.

**Attempt 2 — waive only when the query names a preparation.** The premise was that
prep-qualified queries get their own cache row (`egg fried` and `deviled egg` are rows
distinct from `egg`), so the waiver could reach the funnel's casualty class without
touching the bare keys the golden set asserts on.

The first eval came back green. That was an ordering accident. Probing the decisive case
directly:

```
POST /api/nlp/parse "boiled egg"  -> Boiled Egg | fs_3094 | grams=50
SELECT ... WHERE normalizedForm='egg'  -> egg | Boiled Egg | fatsecret
```

`canonicalBase` strips the preparation, so `boiled egg` collapses onto the bare `egg` row.
Re-running the eval immediately after: **5 real failures, the same egg cases.** One user
typing "boiled egg" is enough to reintroduce the regression.

**So the funnel read's recommendation and the golden set are in direct conflict, and they
are arguing about the same database row.** "boiled egg should cache the FatSecret record"
and "`eggs` must bill ~50g each" cannot both hold while one row serves both queries. No
condition evaluated inside `saveValidatedMapping` can separate them, because by the time
the gate runs, both have already been canonicalized to `egg`.

That makes this a key-granularity question — whether a preparation should get its own
mapping row, and what that costs in cache hit rate — which is a much larger change than a
gate waiver and deserves its own design.

What survives and is worth keeping, if this is revisited:
- `hasPreparationConflict` — the funnel's 4 *correct* blocks were preparation mismatches
  (`poached egg` -> *Fried Egg*), a signal orthogonal to which corpus a record came from.
- The finding that `hasDecisiveBrandContext` is required over a bare `detectedBrand`:
  `detectBrandInQuery("brussels sprouts")` returns "sprouts" (Sprouts Farmers Market).
- The eight-key regression set, which pins the blast radius for any future attempt.

No production change shipped. Both gate runs' cache writes were reverted from a snapshot
and live is verified back at 0 real failures / 12 known issues.

PR #144, opened 2026-07-25
feat(cache): conditional admission for near-miss picks under the save gate

evawlve:
Funnel fix 4 of 5. See `sync-docs/funnel_first_read_2026-07-24.md`.

## What the data said

The F1 instrumentation measured the silent `under_gate` population for the first
time. Of 100 picks blocked by the flat 0.85 confidence gate, **56 were good data
thrown away**, and the population is *clustered* just under the bar — 67 at
≥0.75, 99 at ≥0.60 — not spread along it.

## Why not just lower 0.85

Because the 44 bad ones fall into three detectable classes:

1. **cross-brand substitution** — `buffalo wild wings traditional wings` → Zaxby's, `chipotle burrito` → Pollen + Grace
2. **composite-dish undercount** — `spaghetti and meatballs` → meatballs alone
3. **product-form slip** — `meyer lemon` → lemon syrup, `mac and cheese` → dry raw macaroni

The observation that makes this tractable: **good rows almost always have the
record's brand matching a brand token in the query.** So admission requires a
decisively-named brand that the record carries, plus non-degenerate macros.
That takes class 1 by name, and classes 2/3 are overwhelmingly *unbranded*
queries, so requiring a brand excludes them as a population.

## The part that matters after PR #143

PR #143 was closed because it widened a save gate with no structural bound on
which rows it could touch, and silently repointed the eight most-used generic
keys in the cache.

This one has a bound. `deriveMappingCacheKey` prefixes the cache key with the
brand under **exactly** the `hasDecisiveBrandContext` predicate used here (and
when it declines to prefix, it is because the brand tokens are already in the
key). So an admitted pick's key provably contains a brand token and can never be
the bare generic key — `egg`, `chicken`, `broccoli` — that a golden case asserts
on. **Admission cannot displace those rows because it cannot address them.**

Everything admitted still faces all seven gates in `saveValidatedMapping`; this
only decides what is offered to them.

## Known coverage limits, pinned as tests rather than worked around

Both would require changing predicates that the `brand_mismatch` gate and the
cache key itself depend on — each its own change with its own blast radius:

- **Single-word chain brands are not reached.** `hasDecisiveBrandContext`
  qualifies a single token only next to `BRAND_PRODUCT_CONTEXT_TOKENS`, which is
  a *supplement* vocabulary. So `chipotle burrito` and `wendys frosty` fall out
  here. Multi-word chains (`buffalo wild wings`, `papa johns`) and supplement
  brands (`ryse protein`, `dymatize casein`) are converted.
- **Punctuation-collapsed brands fail token agreement.** `a and w root beer` →
  Root Beer [A&W] was a confirmed-good discard and still is: the first brand
  token `a` doesn't appear in `["root","beer","a&w"]`.

Loosening the predicate for admission alone would forfeit the blast-radius
guarantee above, which is the whole reason this is shippable.

## Verification

- 22 new unit tests, every query/record pair a real row from the 1,158-seed cold warm
- Full suite: 1,082 passing (14 suites + 1 test fail pre-existing on `DATABASE_URL` / `canned tuna in water`)
- `npm run build` clean, typecheck clean
- Box eval gate: pending, results to follow in a comment
- New `mapping.sub_threshold_admitted` log line so the next funnel read can measure this relaxation directly

### DIFF
--- a/src/lib/mapping/__tests__/sub-threshold-admission.test.ts
+++ b/src/lib/mapping/__tests__/sub-threshold-admission.test.ts
@@ -1,0 +1,246 @@
+/**
+ * Conditional sub-threshold admission (funnel fix 4).
+ *
+ * Every query/record pair below is a REAL row from the 1,158-seed cold warm of
+ * 2026-07-24 (sync-docs/funnel_first_read_2026-07-24.md), at the confidence the
+ * cascade actually produced. The blocked cases are as load-bearing as the
+ * admitted ones: the whole point is that lowering the flat 0.85 would let the
+ * cross-brand substitutions in alongside the good picks.
+ */
+
+// simple-rerank reaches the prisma client through its import chain; the
+// admission decision itself touches no database.
+jest.mock('../../db', () => ({ prisma: {} }));
+
+import {
+    assessSubThresholdAdmission,
+    SUB_THRESHOLD_SAVE_FLOOR,
+    SAVE_CONFIDENCE_THRESHOLD,
+} from '../sub-threshold-admission';
+
+const plausible = { kcal: 150, protein: 8, carbs: 20, fat: 4 };
+
+function assess(over: {
+    rawLine: string;
+    confidence: number;
+    matchedBrand?: string | null;
+    isBranded?: boolean;
+    foodName: string;
+    brandName?: string | null;
+    nutrientsPer100g?: { kcal: number; protein: number; carbs: number; fat: number };
+}) {
+    return assessSubThresholdAdmission({
+        rawLine: over.rawLine,
+        confidence: over.confidence,
+        brandDetection: {
+            isBranded: over.isBranded ?? true,
+            matchedBrand: over.matchedBrand ?? null,
+        },
+        foodName: over.foodName,
+        brandName: over.brandName,
+        nutrientsPer100g: over.nutrientsPer100g ?? plausible,
+    });
+}
+
+describe('admits the good picks the gate was discarding', () => {
+    it('dymatize casein -> Elite Casein [Dymatize] at 0.83', () => {
+        const r = assess({
+            rawLine: 'dymatize casein protein',
+            confidence: 0.83,
+            matchedBrand: 'dymatize',
+            foodName: 'Elite Casein',
+            brandName: 'Dymatize',
+            nutrientsPer100g: { kcal: 360, protein: 72, carbs: 8, fat: 4 },
+        });
+        expect(r.admit).toBe(true);
+    });
+
+    it('accepts the brand carried in the NAME rather than the brand field', () => {
+        const r = assess({
+            rawLine: 'ghost protein powder',
+            confidence: 0.8,
+            matchedBrand: 'ghost',
+            foodName: 'Ghost Whey Protein Cereal Milk',
+            brandName: undefined,
+        });
+        expect(r.admit).toBe(true);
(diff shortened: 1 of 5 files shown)


evawlve:
## Box gate: PASSED — and the first design was measured wrong, then replaced

### The first version admitted nothing

The original design required a brand on every admission, on the reasoning that
`deriveMappingCacheKey` prefixes the key under the same `hasDecisiveBrandContext`
predicate, so an admitted pick could never address a bare generic key. The
guarantee was sound. The fix was useless.

Warming **369 real seeds** (69 branded + 300 mixed, from `sync-docs/seed-corpus-2026-07-24/`)
through it produced **zero admissions**. The under-gate population is
overwhelmingly *unbranded*:

```
ground beef 90/10    → Ground Beef 85% Lean 15% Fat   0.78
ground turkey        → 85% lean ground turkey         0.83
canned tuna in water → Fish, tuna, white, canned...   0.84
honey bunches of oats→ Honey Bunches of Oats          0.82
shrimp and grits     → Corn Grits                     0.75
granola with yogurt  → Yogurt parfait with granola    0.77
```

The funnel read's own confirmed-good discards — `ground turkey` → 85% lean
ground turkey, `baked salmon` → Baked or Broiled Salmon — are precisely the rows
a brand requirement excludes.

### What replaced it

The brand requirement is dropped in favour of the guarantee it was standing in
for. The failure that closed PR #143 was **displacement**. These drops are cache
**misses** — by construction there is no incumbent to lose. So an admitted
sub-threshold pick may now **create** a cache row and may **never overwrite**
one (`insertOnly` in `saveValidatedMapping`). A pick the cascade was 78% sure of
cannot evict a row a more confident pick already earned, so no currently-passing
lookup can regress.

Cross-brand substitution is still rejected when the query *does* assert a brand.

### Gate results

| check | result |
|---|---|
| eval before admissions exist | ✅ 0 real failures / 12 known |
| eval, second pass | ✅ 0 real failures / 12 known |
| eval **after** the new rows are cached | ✅ 0 real failures / 12 known |
| eval against live `:3000` (deployed master, shared DB) | ✅ 0 real failures / 12 known |
| **displacements** | **0** |

### Measured yield — 5 admissions on a 300-seed batch

Verified from the database, not from logs (`LOG_LEVEL=info` did not take effect
on the gate server, so the log-based count was unreliable and the DB is the
ground truth):

```
10 15% 85% 90 beef fat lean  → Ground Beef 85% Lean 15% Fat   0.778
15% 85% fat lean turkey      → 85% lean ground turkey         0.831
bunch honey oat of rolled    → Honey Bunches of Oats          0.817
granola with yogurt          → Yogurt parfait with granola    0.768
bang candy cotton            → Cotton Candy                   0.823
```

That is ~5 of the 8 sub-threshold picks in the batch, tracking the funnel's
"56 of 100 were good" estimate. Notably **`shrimp and grits` → *Corn Grits* was
still rejected** by the existing gates — the product-form slip class stopped
without needing a rule of its own.

7 new keys total, 0 identity changes. Snapshot before the run:
`owner@192.168.1.133:/tmp/fm-pre-gate144b.sql`.

### Note on shared-DB state

These 7 rows were written to the production DB by the gate build, as eval/warm
runs always do. They are insert-only additions on keys that previously had no
row, and deployed master would create them identically once this merges. Live
eval is verified green with them present.

PR #145, opened 2026-07-25
fix(fatsecret): a literal "100 g" serving is a panel, not a portion

evawlve:
Funnel fix 5 of 5. **Gated green, but there is a visible trade-off — flagging it for your call rather than merging.**

## What the flat-100g class actually is

Measured on 30 chain queries against the deployed box, **6 still landed at exactly
100g, and every one was a drink**: `starbucks flat white`, `pike place roast`,
`blonde vanilla latte`, `iced matcha latte`, `dunkin matcha latte`,
`dunkin midnight coffee`. The other 24 chain items already resolve to real
differentiated grams — `applebees boneless wings` 330g, `chilis fajitas` 170g,
`ihop french toast` 245g — so fix 1's gram derivation had already cleared most of
this finding.

## Root cause: not a fallback at all

FatSecret ships these records with exactly two servings:

```
fsId 8729727 "Flat White (Tall)"
  "100 g"     grams=100  numberOfUnits=100  ← synthetic per-100g PANEL
  "1 serving" grams=NULL numberOfUnits=1    ← the actual drink, full macros
```

Only the panel row has grams, so it was the only "usable" serving. `grams` was
therefore never null, and the `fs_serving_macros_only` branch that exists
precisely to bill these rows **never ran**. A user logging a flat white was
billed 100 grams of one.

Excluding the panel row from portion selection lets the real serving through. The
filter requires `numberOfUnits === 100` *and* a bare `100 g` description, so a
genuine 100 g product serving is untouched (pinned as a test), and gram-anchored
requests still work — `100g of X` goes through the explicit-weight path, which
does not read this list.

## Second half: the invariant this exposed

Billing a serving from its own macros makes `grams` an energy-density estimate
rather than a weight, so the per-100g block and the billed macros stop agreeing.
The client rescales a portion as `per100g × grams`, so the response carried the
calorie count **twice, with two different answers** — and nudging the portion
would silently switch to the wrong one:

```
flat white:  85g billed 170 kcal, but kcal100 50 × 0.85 = 42
```

The billed macros are authoritative, so per-100g is re-derived from them — the
same correction fix 1 applies to an *absent* panel, extended to a merely
inconsistent one. `per100g × grams == billed` then holds by construction.

## Results

| food | live (`:3000`) | this branch |
|---|---|---|
| starbucks flat white | 100g / **50 kcal** | 85g / **170 kcal** ✅ |
| starbucks pike place roast | 100g / 1 kcal | 2.5g / 5 kcal |
| dunkin matcha latte | 90g / 180 kcal | unchanged |
| applebees boneless wings | 330g / 660 kcal | unchanged |
| quest protein bar | 60g / 170 kcal | unchanged |
| banana | 118g / 105 kcal | unchanged |

170 kcal is the correct value for a tall flat white. The `per100g × grams ==
billed` invariant now holds on every row above, and non-beverage foods are
byte-identical.

- eval: ✅ 0 real failures / 12 known issues
- live re-verified after teardown: ✅ 0 real failures / 12 known
- full suite 1,095 passing (14 suites + 1 test fail pre-existing)

## ⚠️ The trade-off I want your call on

**`pike place roast` displays 2.5g.** For a near-zero-calorie drink the
energy-density estimate is meaningless — a tall brewed coffee is ~355 ml. The
*calories* are right (5 kcal) and the invariant holds, so calorie tracking
improves, but the displayed portion is more obviously wrong than the old 100g.

The real answer for beverages is a **volume** anchor (fl oz parsed from the
serving description, or a drink-size lexicon), not energy density. That is a
separate change and I did not want to invent a gram floor without data to set it.

So: merge this for the calorie correctness and fix beverage grams next, or hold
it until the volume anchor lands and ship them together.

### DIFF
--- a/src/app/api/nlp/parse/route.ts
+++ b/src/app/api/nlp/parse/route.ts
@@ -160,8 +160,10 @@ export async function POST(req: NextRequest) {
     const { forceSegmentText } = await import('@/lib/nlp/heuristic-segmenter');
     const { parseIngredientLine } = await import('@/lib/parse/ingredient-line');
     const { mapIngredientWithFallback } = await import('@/lib/mapping/map-ingredient-with-fallback');
-    const { resolveFoodDetails, isDegenerateNutrition, per100gFromBilledMacros } =
-      await import('@/lib/nlp/resolve-payload');
+    const {
+      resolveFoodDetails, isDegenerateNutrition, per100gFromBilledMacros,
+      isPer100gInconsistentWithBilled,
+    } = await import('@/lib/nlp/resolve-payload');
     const { logger } = await import('@/lib/logger');
 
     const body = await req.json();
@@ -347,7 +349,18 @@ export async function POST(req: NextRequest) {
       // persisted by a background task — that split shipped correct serving
       // calories next to kcal100: 0, and the client rescales by kcal100. Fall
       // back to the line's own macros so the two can't contradict each other.
-      const derivedPer100g = isDegenerateNutrition(details.nutritionPer100g)
+      //
+      // The same split reappears whenever the two are merely INCONSISTENT
+      // rather than absent (funnel fix 5). A record billed from its own
+      // per-serving macros — FatSecret's "1 serving" restaurant rows — carries
+      // grams that are only an energy-density estimate, so a tall flat white
+      // bills its true 170 kcal while kcal100 x grams says 42. Whichever number
+      // the client happens to use then decides the calorie count, and changing
+      // the portion silently switches it to the wrong one. The billed macros are
+      // authoritative, so re-derive per-100g from them and the invariant
+      // per100g x grams == billed holds by construction at any portion.
+      const inconsistent = isPer100gInconsistentWithBilled(details.nutritionPer100g, mapped);
+      const derivedPer100g = isDegenerateNutrition(details.nutritionPer100g) || inconsistent
         ? per100gFromBilledMacros(mapped)
         : null;
       const nutritionPer100g = derivedPer100g
--- a/src/lib/mapping/__tests__/build-fatsecret-result.test.ts
+++ b/src/lib/mapping/__tests__/build-fatsecret-result.test.ts
@@ -385,3 +385,96 @@ describe('repro: "15 pretzels" (fs_4349 live data, eval n-serv-20 675g regressio
         expect(result!.servingTier).toBe('fs_label_count');
     });
 });
+
+// Funnel fix 5. A literal "100 g" serving is a per-100g PANEL, not a portion.
+// FatSecret ships chain-restaurant records with exactly two servings: the
+// synthetic `100 g` row and the real `1 serving` row that carries the item's
+// macros but no gram weight. Because only the former had grams it was the only
+// "usable" one, so `grams` was never null, the macro-only branch never ran, and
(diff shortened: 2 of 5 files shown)


evawlve:
## Update — the trade-off is resolved, no volume anchor needed

Pushed `753fe5e`. **`pike place roast` now reads 477g, not 2.5g**, so the thing
blocking this merge is gone. It did not need a drink lexicon or an invented gram
floor.

### The 2.5g was never a beverage problem

Grams came from `PREPARED_FOOD_KCAL_PER_GRAM = 2.0` — a prepared-**solid**
density. 170/2.0 = 85g and 5/2.0 = 2.5g reproduce both numbers in the table
above exactly. Energy density is the wrong instrument, and it's wrong for every
record on this path, not just drinks.

### The record already carries the answer

FatSecret derives the synthetic `100 g` panel FROM the serving
(`serving / grams × 100`), so the division is **reversible**:
`100 × serving_nutrient / panel_nutrient` returns the grams that produced it.

Checked against the 24,579 servings on the box whose true weight IS recorded:

| | ratio predicted/true |
|---|---|
| median | **1.000** |
| p10 / p90 | 0.995 / 1.005 |
| within 5% | **98.1%** |

That's arithmetic, not estimation.

### Median across nutrients, because the panel is rounded

Relative quantization error grows as the value shrinks — a `1 kcal/100g` brewed
coffee is ±50% before anything else goes wrong. Measured, share landing within
10% of true weight:

| panel kcal/100g | calories only | median-of-nutrients |
|---|---|---|
| ≤5 (drinks) | 68.8% | **91.1%** |
| 6–20 | 92.9% | **99.0%** |
| >20 (solids) | 99.7% | **99.9%** |

Only nutrients positive on *both* sides vote, so a serving fat rounded down to 0
against a positive panel fat can't drag the median toward zero.

### It also stops making the zero-calorie drinks worse

For a genuine 0 kcal record — Diet Coke, sparkling water, unsweetened tea, 50 of
them on the box — every panel nutrient is 0, so there is nothing to invert.
Demoting the panel there traded a 100g display for a **1g** one showing the same
correct 0 kcal. Those now keep today's behaviour untouched.

### Live gate (box `/tmp/gate146` on :3100 vs deployed :3000)

| query | live | this branch |
|---|---|---|
| starbucks flat white | 100g / 50 kcal | **340g / 170 kcal** |
| starbucks pike place roast | 100g / 1 kcal | **477g / 5 kcal** |
| starbucks blonde vanilla latte | 100g / 53 kcal | **227g / 120 kcal** |
| impossible whopper | 270g / 675 | unchanged |
| applebees boneless wings | 330g / 660 | unchanged |
| quest protein bar | 60g / 170 | unchanged |
| banana | 118g / 105 | unchanged |
| chicken breast | 112g / 130 | unchanged |

`per100g × grams == billed` verified on every row, both sides.

- eval gate: ✅ **0 real failures / 12 known** (exit 0); same 3 known-issue
  promotions as the last FS run (`s-supp-11`, `n-serv-57`, `n-mq-22`)
- FoodMapping diffed before/after the gate run: **0 added, 0 removed, 0 changed**
- covers 1,520 of the 1,576 panel-only records — not only the 6 drinks
- records with no panel at all still use the energy-density estimate, which is
  what it was written for

PR #154, opened 2026-07-25
feat(eval): F2 failure-class registry + F3 warm-and-triage driver

evawlve:
## F2 — the registry

`scripts/eval/failure-classes.ts`: **32 typed classes — 18 pipeline / 8 healthy / 3 data / 3 ingest** — seeded from the 19 `dropReason` classes actually observed on the box, the 13 cold classes from the 2026-07-24 funnel read, and the 296-defect 2026-07-21 triage vocabulary. Plus `classify-drops.ts` (warm JSONL **or** `MappingEventLog`, read-only) and a generated `sync-docs/failure-classes.md`.

### Four measured findings shaped it, and each would have produced a wrong registry

**1. The raw histogram lies.** `save_rejected:cross_source_margin` is 404 of 571 `save_rejected` events — **71%** — and it is **incumbent protection**, not a defect. The gate sits inside `if (newTargetKey && existingTargetKey && ...)` (`validated-mapping-helpers.ts:724`), so on a cold key `existing` is null and the block is skipped: **it cannot fire without an incumbent.** A registry ranked by raw count sends its first and largest fix-agent at correct code. Hence `fixKind: 'healthy'`, `status: 'by-design'`, and counts split incumbent-vs-cold.

**2. You cannot join event rows to cache rows on `normalizedForm`.** `MappingEventLog.normalizedForm` is *pre*-canonicalization; `FoodMapping`'s is canonicalized, singularized and token-**sorted**. `chicken breast` → `breast chicken`, `eggs` → `egg`, `pretzels` → `pretzel`. A naive join reports cached keys as cold and **inverts the conclusion** — it did exactly that during scoping. `canonicalizeCacheKey` is imported from the real module and applied at a single chokepoint.

**3. `under_gate`'s dropReason is not diagnostic.** F1 tags it with the confidence gate's `selectionReason` — which path *selected* the pick — and `simple_rerank` / `close_match` / `clear_winner` each split ~50/50 good-vs-bad. Classes under `under_gate` are distinguished by row heuristics, never by that string.

**4. The funnel measures conversion, not correctness.** 38 all-zero-macro rows counted as `saved` — successes. A quality dimension sits alongside the stage so "converted, but poisoned" is expressible.

### `fixKind` is the load-bearing addition

Only `'pipeline'` classes may ever be handed to a fix-agent. `'ingest'` means no record exists in any lane and **no code change produces the right number**. `'data'` means a repoint batch. `'healthy'` is counted so it stops reading as a defect.

Detectors carry both `exemplarRows` **and `counterRows`** — rows they must *not* match — so a passing test cannot be tautological.

### Verified against the live box (read-only, 2,000 rows since 2026-07-24)

```
rows=2000  defect events=561  healthy=1439  unclassified=0 (0.00%)

DEFECTS (ranked by DISTINCT KEYS, not events)
  identity-query-token-dropped     38 keys  323 events  pipeline
  identity-brand-substituted        9        100        pipeline
  composite-component-drift         8         66        pipeline
  serving-flat-100g-fallthrough     6         49        pipeline  fixing
HEALTHY / BY DESIGN
  gate-cross-source-margin          8         57        by-design
  gate-sub-threshold-no-displace     3         21        by-design

Live-dropReason coverage: all 19 measured classes owned.
```

## F3 — the driver

`scripts/eval/triage-drops.ts` wires drops through the probes into one dossier per key, classifies via F2, and emits `triage-verdicts-<ts>.json` + `.md` **keeping every key of the 2026-07-21 report shape** (`screened` / `suspectCount` / `confirmedCount` / `byClass` / `confirmed` / `dismissed`, and `seed` / `confirmed` / `cls` / `action` / `targetId` / `note` per row) so existing readers still work.

**The sprint spec called F3 "pure wiring of scaffolding that already exists". That was false.** `debug-retrieval-probe.ts`, `filter-trace-probe.ts`, `full-pipeline-probe.ts`, `warm-names.ts` and `golden-diff-eval.ts` existed only in the Syncthing tree and were **never committed**. They are now in the repo and importable as functions while keeping their CLI behaviour.

**Nothing is silently truncated.** The run logs its stage filter, its dedupe, its per-class cap and whether the cap dropped anything, and states plainly that `--no-probes` verdicts rest on row facts alone.

## Two limits named, both found by running it rather than reading it

- **`hadIncumbent` is wrong in *both* directions.** Resolved at report time, so a key cached after the event reads true *and* a key evicted after reads false. The first run reported 8 "cold" `cross_source_margin` events — impossible for that gate. They were rows the composite cache refresh had evicted hours later. A small cold count there is a timing artifact; a large one is a real signal.
- **`cacheKey` is the READ key, not always the WRITE key.** `deriveMappingCacheKey` brand-prefixes under decisive brand context. Measured: for every branded query tried the two agree, because the event log's `normalizedForm` already carries the brand token — but the log records neither the detected brand nor the decisive-context verdict, so where the save path did prefix, this lookup cannot reproduce that key.

## Gate

- **1,733 tests across 94 suites**, `tsc` clean, `lint:ci` **0 errors**
- `classify-drops` and `triage-drops` both exercised against the live event log, read-only, no DB writes

## Next

F4 — the per-class fix workflow — is written and parked at `sync-docs/warm-plan-2026-07-25/f4/` in the mobile repo, pending this merge. Diego's standing decision holds: report-only nightly, fix-agents on manual trigger only.

### DIFF
--- a/scripts/debug-retrieval-probe.ts
+++ b/scripts/debug-retrieval-probe.ts
@@ -1,0 +1,136 @@
+/**
+ * debug-retrieval-probe.ts — for each query, dump what EACH retrieval source returns.
+ *
+ * Answers "why did this line drop?" by separating a DATASET gap (nothing exists in
+ * any lane) from a LANE/RANKING gap (records exist, retrieval or rerank lost them).
+ *
+ * ---------------------------------------------------------------------------
+ * COST + SIDE EFFECTS (read this before running it over hundreds of drops)
+ * ---------------------------------------------------------------------------
+ * - NO LLM CALLS. Nothing here reaches OpenRouter/Ollama. `gatherCandidates` does
+ *   keyword search (Typesense `off_foods` + Postgres FDC) and, only when
+ *   SEMANTIC_SEARCH_ENABLED=true, a LOCAL ONNX query embedding — no paid tokens.
+ * - IT IS NOT PURELY READ-ONLY. `searchFatSecretLane` fires
+ *   `persistFatSecretHits` as a registered background task, which UPSERTS
+ *   FatSecretFood / FatSecretServing rows (corpus hydration of FS API results —
+ *   it never touches FoodMapping, so it cannot change a cached pick). For a
+ *   strictly read-only run set FATSECRET_RETRIEVAL_ENABLED=false BEFORE this
+ *   module is loaded; the lane then returns [] without calling the API.
+ * - FatSecret API quota IS consumed per query (Premier tier), twice per query in
+ *   fact: once by the direct lane call and once inside gatherCandidates. Pass
+ *   `{ skipFatSecretLaneCall: true }` to drop the direct call and rely on the
+ *   gather pass alone.
+ *
+ * Usage (unchanged):
+ *   ts-node ... scripts/debug-retrieval-probe.ts '<query1>' '<query2>' ...
+ *   ts-node ... scripts/debug-retrieval-probe.ts --file <queries.txt>
+ */
+import * as nodeFs from 'fs';
+import { searchFatSecretLane } from '../src/lib/mapping/fatsecret-lane';
+import { gatherCandidates } from '../src/lib/mapping/gather-candidates';
+
+/** One candidate, flattened to the fields a triage reader needs. */
+export interface ProbeCandidate {
+    /** `off_<barcode>` | `fdc_<id>` | `fs_<fsId>` — the repoint target vocabulary. */
+    id: string | null;
+    source: string | null;
+    name: string | null;
+    brand: string | null;
+    score: number | null;
+}
+
+export interface RetrievalProbe {
+    query: string;
+    /** Direct FatSecret lane call. `null` when skipped, `{error}` when it threw. */
+    fatsecret: { count: number; top: ProbeCandidate[] } | { error: string } | null;
+    /** Combined OFF + FDC + FS through gatherCandidates (branded path forces OFF on). */
+    gather: { total: number; bySource: Record<string, number>; top: ProbeCandidate[] } | { error: string };
+}
+
+export interface RetrievalProbeOptions {
+    /** How many candidates per source to ask for (default 8, same as the pipeline). */
+    maxPerSource?: number;
+    /** How many candidates to keep in the report (default 6). */
+    topN?: number;
+    /** Skip the direct FatSecret lane call — halves FS API quota per query. */
+    skipFatSecretLaneCall?: boolean;
+}
+
+function toProbeCandidate(c: unknown): ProbeCandidate {
+    const r = c as { id?: string; source?: string; foodName?: string; name?: string; brandName?: string | null; score?: number };
+    return {
+        id: r.id ?? null,
+        source: r.source ?? null,
+        name: r.foodName ?? r.name ?? null,
+        brand: r.brandName ?? null,
+        score: typeof r.score === 'number' ? +r.score.toFixed(3) : null,
+    };
+}
+
+/**
+ * Probe every retrieval lane for one query. Never throws: per-lane failures are
+ * captured as `{ error }` so a batch caller keeps going.
+ */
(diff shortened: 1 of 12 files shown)


evawlve:
## Review pass: 7 blocking findings from two adversarial reviews, all addressed

CI was already green on the first push, which is the point — **none of these would have been caught by it.**

| # | finding | outcome |
|---|---|---|
| **A** | `triage-drops` could emit `targetId: 'fs_<id>'`, and `apply-repoints.ts:83-104` branches only on `off_` before falling through to `parseInt(target.slice(4))` as an fdcId | **fixed** — targets constrained to the vocabulary apply-repoints can consume; anything else emits `null`, names the record in the note, and is counted in the log. A test sweeps all 31 classes × 6 id shapes. |
| **B** | `--concurrency 0` built `Array.from({length: NaN})` → an empty worker array → **zero probes ran while the report said `mode='probes'`** | **fixed** — flags validated, clear error, exit 2 |
| **C** | `--limit` on `--in` truncated silently; every downstream count described the slice, not the file | **fixed** — violated this file's own "TRUNCATION IS ALWAYS REPORTED" contract |
| **D** | "six defect classes and the whole `serving` bucket are unreachable" | **not a defect — premise was wrong.** 5 of the 6 declare `alsoStages` including `under_gate`, which *is* in `DROP_STAGES`. Proven by test and end-to-end. Only the 4 stage-exclusive classes and the `evict` action are genuinely out of reach. Scope disclosure is now **computed** from `DROP_STAGES × registry` rather than asserted. |
| **E** | `cacheKey` was not the FoodMapping key — and the gap was bigger than the brand prefix | **fixed** — real misses: `3 egg whites` `egg`→`egg white`, `1 cup cooked oats` `oat`→`cooked oat`, `rolled rolled oats` `oat rolled rolled`→`oat rolled` |
| **F** | `corrupt-panel-served` was **dead** in `--from-db` (needs a per-100g panel `MappingEventLog` doesn't store) | **fixed + declared** — billing is linear so `100 × totalKcal / grams` inverts the panel exactly; the Atwater arm genuinely can't work and is now surfaced as a declared blind spot |
| **G** | `venue-brand-blackout-tripwire` claimed `no_candidates` and shadowed `ingest-gap-no-candidates` | **fixed** — a filter blackout is by definition `all_filtered` |

### A note on E, because it required a `src/` change

Two leaf-safe steps moved to **`src/lib/mapping/cache-key-core.ts`** (verbatim; `deriveCacheKeyName`'s body diffs byte-identical) and are re-exported from `cache-key.ts`, so the public surface is unchanged. The move is necessary because `cache-key.ts` transitively loads `config.ts` — which snapshots `FATSECRET_RETRIEVAL_ENABLED` and warms ONNX — and that would break the eval tooling's read-only guarantee. An **import-graph test** now pins it, and was mutation-tested: re-adding the `cache-key.ts` import makes it fail.

### And a correction to something this branch had already committed

The earlier commit claimed the 8 "cold" `cross_source_margin` events were an eviction-timing artifact. **That was wrong** — a plausible story that stopped the search early. They were the `cacheKey` bug in E, and fixing it took the count **8 → 0 cold**, exactly as reading the gate predicts. When a structural claim and the data disagree, suspect the measurement.

### Gate

- **1,859 tests across 92 suites** (was 1,733), `tsc` clean, `lint:ci` **0 errors**
- `classify-drops` re-run against the live event log, read-only: **0 unclassified**, all 19 measured `dropReason` classes owned
- The report now carries a **`WHAT THIS INPUT CANNOT SHOW`** block and a **`KEY FIDELITY`** block — 74 rows needed a discriminator to reach their real key, 162 name `white/yolk/cooked/whole`

### Left alone deliberately (non-blocking, listed by the reviewers)

`kcalPer100g>250` over-match; `identity-brand-substituted` missing a query-names-a-brand precondition; composite calibration untested on soup/salad/pizza; `SERVED_STAGES` omitting `save_rejected`; `readInputFile`'s silent 0-row parse. Each is a real observation and none changes a shipped verdict.

PR #157, opened 2026-07-25
fix(mapping): apostrophe-insensitive brand matching

evawlve:
## What

`candidateMatchesTargetBrand` tokenizes on `[\s,()[\]{}]+` — **the apostrophe is not a delimiter** — then asks `candTokens.includes(brandTokens[0])`. A record branded `Member's Mark` yields the token `member's`; the brand lexicon's detected form yields `members`. They never matched.

Found by **warm batch 01** of the seed-corpus campaign: `members mark trail mix` was rejected by the save gate against a record whose brand *is* Member's Mark.

## Why it matters more than the save gate

The save gate is the smaller half. The same function has **five call sites**, and the other three are ranking:

| site | effect when it wrongly returns false |
|---|---|
| `simple-rerank.ts:1672` | loses the **decisive-brand boost** |
| `simple-rerank.ts:1859` | loses same-brand variant precision |
| `sub-threshold-admission.ts:126` | candidate not admitted |
| `validated-mapping-helpers.ts:489,514` | save rejected → `winner = null` → AI backfill at 100 g |

So an affected brand was losing its *ranking* anchor, not just a save.

## Fix

Fold `'` / `’` / `` ` `` out of **both** sides before the whole-token comparison.

This is an **admit-only relaxation** — dropping characters can only merge token forms, so it can only turn a failed match into a successful one. Every one of the five consumers therefore becomes more permissive, never stricter. Top of the preference order in `sync-docs/mapping_investigation_playbook.md`.

**No migration needed** — the function is computed at request time and never persisted.

Deliberately *not* fixed here: loosening the save gate itself. A hard reject there sets `winner = null` and falls through to AI backfill at 100 g, which is worse than the drop.

## Measured scope

Live probe through the real `detectBrandInQuery` → `hasDecisiveBrandContext` → `candidateMatchesTargetBrand` chain — **not** a static join:

| | tokens | corpus records exposed |
|---|---:|---:|
| multi-word lexicon brand vs possessive-first record | **27** | **6,122** |

Nature's Promise 2587 · Member's Mark 1202 · President's Choice 1104 · Boar's Head 397 · Farmer's Union 163 · Baker's Life 153 · Lowe's Foods 76 · Wegman's 65 · Bird's Eye 15.

Two corrections worth recording:

- An earlier **static lexicon × corpus join predicted ~17,200 records and was overstated ~3×.** Single-token possessives (`hersheys`, `campbells`, `lays`, `mcdonalds`, `reeses`, `welchs`, `smuckers`) are detected but score `decisive=false`, so the gate never fires and they are admitted normally. Only multi-word brands, where `hasDecisiveBrandContext` short-circuits true, are bitten.
- **6,122 is the exposed record population, not the defect count.** The gate only fires when such a record is the winner for a decisive-brand query.

Probe-confirmed rejections: `boars head turkey breast`, `members mark trail mix`, `bobs red mill oats`. Controls `great value` / `kirkland signature` admit correctly both before and after.

## Relationship to #149

Same root cause, different path. #149 fixed apostrophes in the **cache key**; this fixes apostrophes in the **brand comparison**.

A third, separate class remains open and is **not** in this PR: where the lexicon carries *only* the possessive form, an apostrophe-free query (`chilis chicken crispers`, `applebees riblets`) returns `matchedBrand: null` from `detectBrandInQuery`. That is a lexicon **data** fix (add apostrophe-free aliases) and is being sized separately.

## Verification

- New `possessive-brand-match.test.ts`: **5 of 7 fail** against the unfixed function, **7 of 7 pass** with it. The two negative controls — unrelated brand (`Udi's` vs `great value`), and the substring hazard (`one` must not match `Toblerone`) — pass in *both* directions, so folding has not silently become substring matching. A degenerate all-apostrophe target is rejected rather than matching everything.
- Full `src/lib/mapping` suite: **774 pass**. One pre-existing failure, `does NOT fast-path "canned tuna in water"`, **also fails on clean master** and is unrelated (CI's `build` job does not run jest).
- `lint:ci` 0 errors · `typecheck` clean.
- Box eval gate to be run against this branch before merge.

### DIFF
--- a/src/lib/mapping/__tests__/possessive-brand-match.test.ts
+++ b/src/lib/mapping/__tests__/possessive-brand-match.test.ts
@@ -1,0 +1,78 @@
+import { candidateMatchesTargetBrand, simpleRerank, type RerankCandidate } from '../simple-rerank';
+
+/**
+ * Possessive brands defeat the whole-token brand matcher (found by warm batch 01,
+ * Jul 2026).
+ *
+ * `candidateMatchesTargetBrand` tokenizes on `[\s,()[\]{}]+`, which does not
+ * include the apostrophe, so a record branded "Member's Mark" yields the token
+ * `member's` while the lexicon's detected form yields `members`. They never
+ * matched. Because the function is consulted by the rerank decisive-brand boost,
+ * same-brand variant precision, sub-threshold admission and the save gate, a
+ * possessive-first brand lost its *ranking* boost as well as its save — the
+ * larger of the two effects.
+ *
+ * Measured exposure at the time of the fix: 27 multi-word lexicon brands against
+ * 6,122 corpus records — Nature's Promise (2587), Member's Mark (1202),
+ * President's Choice (1104), Boar's Head (397).
+ *
+ * Same root cause as the PR #149 cache-key fix, on a different path: #149 was
+ * apostrophes in the key, this is apostrophes in the brand comparison.
+ */
+
+function cand(partial: Partial<RerankCandidate> & { id: string; name: string }): RerankCandidate {
+    return { score: 0.5, source: 'openfoodfacts', ...partial };
+}
+
+describe('candidateMatchesTargetBrand — apostrophe folding', () => {
+    it('matches an apostrophe-free lexicon brand against a possessive record brand', () => {
+        // The batch-01 defect verbatim: `members mark trail mix` was rejected
+        // against a record whose brand is literally "Member's Mark".
+        expect(candidateMatchesTargetBrand('Member\'s Mark', 'Trail Mix', 'members mark')).toBe(true);
+        expect(candidateMatchesTargetBrand('Boar\'s Head', 'Ovengold Turkey Breast', 'boars head')).toBe(true);
+        expect(candidateMatchesTargetBrand('Nature\'s Promise', 'Organic Baby Spinach', 'natures promise')).toBe(true);
+    });
+
+    it('matches in the reverse direction: possessive lexicon form, bare record brand', () => {
+        expect(candidateMatchesTargetBrand('Members Mark', 'Trail Mix', 'member\'s mark')).toBe(true);
+        expect(candidateMatchesTargetBrand('Boars Head', 'Ovengold Turkey Breast', 'boar\'s head')).toBe(true);
+    });
+
+    it('folds the typographic apostrophe too — OFF carries both forms', () => {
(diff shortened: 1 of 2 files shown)


evawlve:
## Box gate: PASS — but with one caveat that narrows the claim above

Isolated gate per the documented recipe (branch built in `/tmp/gate157`, served on `:3100`; deployed `recipe-api` on `:3000` never touched).

| | result |
|---|---|
| eval vs branch | **0 real failures / 15 knownIssue** — identical to the post-batch-01 baseline |
| FoodMapping diff across the eval | **0 added / 0 changed / 0 removed** |
| live `:3000` after gate teardown | healthy |

### Live behaviour, before → after

| query | deployed master | this branch |
|---|---|---|
| `members mark trail mix` | `save_rejected:brand_mismatch` | **saved** — *Trail Mix*, Member's Mark (fatsecret) |
| `boars head turkey breast` | dropped the brand boost | **saved** — *Boars Head Mesquite Turkey* |

### The caveat — an over-admission this fix newly enables

`natures promise baby spinach` now **saves**, mapped to *"Nature's Promise, Organic Baby Food, Pear, Peas & Spinach"* — infant purée, not baby spinach leaves.

That pick was **already the winner before this change** and already served the user; the brand gate only controls whether it gets *cached*. So the possessive bug was accidentally acting as a filter on wrong-identity picks for possessive brands specifically.

This does not make the change unsafe, but it makes the "admit-only ⇒ safe" claim in the description too broad, so stating it precisely:

> Admit-only guarantees the change can never *reject* something it previously accepted. It does **not** guarantee the newly-admitted rows are correct — it converts pre-existing wrong picks from transient into cached.

The protection being lost was coincidental, not designed: it only ever covered possessive brands, never `great value baby spinach`. The real defect is retrieval ranking a purée above leaf spinach, which is a separate, pre-existing issue.

The offending row was deleted from the live cache (`DELETE 1`, verified). The two correct probe rows were kept. Net cache impact of this gate run: **+2 rows, both correct.**

### Methodology note worth recording

The first eval run against "this branch" was **void and looked green**. Port `:3100` was already held by a `next-server` left over from PR #152's gate run seven hours earlier; my server hit `EADDRINUSE` and died, and the `curl /api/ok` health check happily passed against the squatter. The result was old code wearing the branch's name.

It was caught only because a live probe disagreed with the same function called locally. **A health check is not an identity check** — verify the build by process `cwd` (`readlink /proc/$PID/cwd`), and check for `EADDRINUSE` in the server log before trusting any gate result.

PR #160, opened 2026-07-26
fix(mapping): possessive brands survive the required-token filter

evawlve:
Third apostrophe fork in this pipeline, after the cache key (#149) and the rerank brand matcher (#157). Same root cause, third code path — this one in the **filter**, upstream of everything #157 touched.

## The defect

`tokenize()` splits on `[^\w]+`. The apostrophe is not a word character, so `"Trader Joe's"` yields `{trader, joe}` and the orphan `s` is dropped by the `length > 2` filter. A brand-leading query normalizes to `trader joes`, whose must-have tokens are `["trader","joes"]`, and `joes` matches nothing. The plural rescue can't bridge it either — `singularize('joes')` hits the `-oes` branch and returns `jo`.

**This annihilates the pool rather than demoting it.** From the warm batch 01 filter log:

```
filter.candidates.removed normalizedName="trader joes scandinavian swimmers"
mustHaveTokens=["trader","joes"] before=12 after=1
removed=["Scandinavian Swimmers","SUPER SOUR SCANDINAVIAN SWIMMERS","Scandinavian Swimmers"]
```

The only survivors are records whose brand string happens to omit the apostrophe — whatever food they are. Three of the four affected rows reached rerank as `pool=1 single_candidate`. The reranker did not choose wrong; it was handed one option.

## The design decision worth reviewing

The fold is **additive**, and that is the entire point:

- folded tokens are **added to** the token set, not substituted for it
- the folded name is tested **alongside** the raw name, not instead of it

So the token set is a superset of today's and the name tests are a disjunction over today's. **No candidate that survives the required-token check today can fail it now — by construction, not by measurement.**

The obvious alternative — fold in place — is wrong, and this was measured independently before shipping: substituting collapses `"Wendy's"` to the single token `wendys`, losing the `wendy` token that a possessive-spelled query needs, and `pluralize('wendy')` is `wendies` so nothing bridges it. **Candidate-side substitution drops 5 of 22 possessive-brand pairs that pass today** (`wendy's`, `arby's`, `denny's`, `zaxby's`). Those four are in the test file's last block as survivors: they pass *before and after* this change.

## Tests

`src/lib/mapping/__tests__/possessive-brand-token-filter.test.ts` — 13 tests.

- **5 fail without the fix, 13 pass with it.**
- 8 pass in both directions and are deliberate controls: the two substring-hazard negatives (wrong brand, `one`/`Toblerone`), the no-apostrophe fast path, and the four substitution survivors above.
- The first test asserts the *mechanism* (`deriveMustHaveTokens('trader joes scandinavian swimmers') === ['trader','joes']`) so that if the derivation changes, this fix's premise fails loudly rather than silently becoming a no-op.

`npm run typecheck` clean. `npx jest src/lib/mapping` → 799/800; the one failure (`does NOT fast-path "canned tuna in water"`) is pre-existing on clean master — an `aiGeneratedServing` mock gap, unrelated.

## What this does NOT fix

- **`singularize()`'s `-oes` branch still returns `jo` for `joes`.** The fold makes it unnecessary on this path. It is one of the three cruder shadow singularizers #151 flagged and left open.
- **Admit-only is not the same as correct.** This is the #157 lesson repeated: opening these pools lets the reranker answer `trader joes gone bananas` with *Dried Baby Bananas* instead of a chicken breast — a better food, still the wrong one, and now cacheable. Four batch-01 rows flip to the right record under this change; four more move to a better-but-still-wrong record. The ranking clusters behind those four are separate work.

## Gate

Golden cases (`n-mq-43`..`n-mq-47`, including the `trader joes chicken breast` over-admission tripwire) are being measured against the live corpus on an isolated build and will be added to this branch from the measurement, not written ahead of it.


evawlve:
## Live measurement on the deployed corpus

Method: two isolated trees on the box (`/tmp/gate160base` = deployed master `e5a49446`, `/tmp/gate160` = this branch), staged from the same `node_modules` and `.env`, verified to differ **only** in `filter-candidates.ts`. Every seed resolved **cold** (`skipCache: true`) through the real `mapIngredientWithFallback` — not a probe that reimplements the caller's argument list. Prisma writes blocked by a guard (2,132 write calls blocked, incl. 72 `foodMapping.upsert`), so nothing touched production data.

### The pool evidence

This is the finding, more than any individual flip. Rerank pool size, before → after:

```
bagel but everything joe seasoning the trader   1 -> 4
banana gone joe trader                          1 -> 10
brew cold joe trader                            1 -> 10
burger joe trader turkey                        1 -> 8
cheddar joe trader unexpected                   2 -> 10
chicken mark member rotisserie                  2 -> 10
dip elote joe trader                            1 -> 10
joe scandinavian swimmer trader                 1 -> 10
almondmilk joe trader vanilla yogurt            1 -> 10
burrito chicken joe kung pao trader             1 -> 10
joe pasta protein trader                        2 -> 10
```

Eleven seeds were reaching the reranker with **one or two candidates**. The reranker was not choosing badly; it was being handed a single option and asked to confirm it.

### Outcome on the 17-case group-A set

**base 1/17 correct → patched 6/17.** No case regressed.

| flip | before | after |
|---|---|---|
| `trader joes cold brew` | Cold Pressed Juice, Red | **Black Cold Brew Coffee** [Trader Joe's] 240 g / 10 kcal |
| `trader joes turkey burgers` | Trader Joe's Turkey **Meatballs** | **Turkey Burgers** [Trader Joe's] 113 g / 180 kcal |
| `trader joes unexpected cheddar` | Unexpected cheddar **chicken sausage** | **Unexpected Cheddar Cheese** [Trader Joe's] 28 g / 110 kcal |
| `trader joes scandinavian swimmers` | Paper Thin Scandinavian **Crackers** | **Scandinavian Swimmers** [Trader Joe's] 28 g / 100 kcal |

A fifth case, `trader joes kung pao chicken burrito`, satisfies its `/kung pao/i` assertion (King Oyster Mushrooms → *Low Fat Kung Pao Chicken Stir Fry*) but I am **not counting it as fixed**: the seed is a burrito and the record is a stir fry at 120 g / 240 kcal. That is the composite→component class, unrelated to this PR.

Four more moved to a better-but-still-wrong record (`gone bananas` → Dried Baby Bananas, `elote dip` → Tzatziki Dip, `vanilla yogurt` → Organic Vanilla Yogurt, `protein pasta` → Orzo Pasta). Those are the ranking clusters, and they are the reason the campaign's correctness screen has to land before batch 02 — this change makes them *cacheable*.

### Hold-set 1 — all 81 labelled batch-01 seeds, cold, both trees

```
GOOD      1 of 48 moved
SUSPECT   0 of 10 moved
BAD      10 of 23 moved
```

The single GOOD move is lateral and I checked it rather than waving at it: `trader joes tzatziki` goes from `fs_69274` *Tzatziki Dip* [Trader Joe's] 30 g / 35 kcal @ conf 0.76 `simple_rerank` to `off_0002000496937` *Tzatziki* [Trader Joe's] 28 g / 35 kcal @ conf 0.88 `branded_exact_match`. Same brand, same food, same calories, and the new pick is an exact branded match rather than a generic rerank result. The pool count stayed at 10 because the newly-admitted OFF record displaced another candidate inside the top-10 slice.

### Hold-set 2 — the pairs the alternative implementation would have broken

**0 of 12 changed**, byte-identical winners:

`wendys spicy chicken sandwich` · `arbys roast beef sandwich` · `dennys pancakes` · `zaxbys chicken fingers` · `boars head ovengold turkey breast` · `natures promise baby spinach` · `members mark trail mix` · `udis gluten free bread` · `jack in the box curly fries` · `mcdonalds big mac` · `presidents choice granola` · `trader joes chicken breast`

The first four are the specific pairs that a *substituting* candidate-side fold drops. This is the empirical half of the argument the commit message makes structurally.

### Two things this measurement found that are not this PR

1. **`trader joes chicken breast` answers with Mary's Chicken** — `off_0208065711897`, 112 g / 130 kcal — while `off_0274038708522`, the real Trader Joe's record, exists. Right food, wrong brand, right macros, so every numeric band passes and only the brand is wrong. Pinned as `n-mq-47` with `knownIssue: true`.
2. **`presidents choice granola` throws** in *both* trees: `Invalid prisma.aiGeneratedServing...`. Same root cause as the one pre-existing jest failure on master (`does NOT fast-path "canned tuna in water"`). Identical before and after, so not a regression here, but it is a real crash on a real query.

### Golden cases

`n-mq-43`..`n-mq-47` added in 6b22e47, every value taken from the measurement above. `n-mq-47` needs a `--write-baseline` refresh after deploy.

evawlve:
### Correction to the measurement comment — finding 2 was mine, not the code's

I reported that `presidents choice granola` throws `Invalid prisma.aiGeneratedServing.findUnique()` on a real query. **That is a probe artifact, not a live defect.** Retracting it.

The probes install a hard write-guard (`scripts/probe/guard.ts`) that stubs every mutating delegate:

```ts
d[op] = async () => { note(`${m}.${op}`); return { count: 0 }; };
```

`presidents choice granola` has no corpus record, so it falls through to the AI nutrition backfill. `saveAiNutrition` creates an `AiGeneratedFood` row and returns `record.id` — but under the guard the create returns `{ count: 0 }`, so `record.id` is `undefined`, `saveAiNutrition` returns `undefined` despite being typed `Promise<string>`, and `getAiServingGrams(undefined, …)` throws at `ai-nutrition-backfill.ts:609`. In production the create succeeds and `foodId` is a real string.

The same mechanism explains the pre-existing jest failure I cited alongside it (`does NOT fast-path "canned tuna in water"`, `ambiguous-unit-backfill.ts:48`) — that one is a *mock* gap rather than a guard stub, but it is the same "the write did not really happen, so the id is undefined" shape. Neither is evidence of a production crash.

**Nothing else in the measurement changes.** The seed threw identically in both trees, so it contributed nothing to the before/after diff, and every other row resolved normally. The pool sizes, the 6/17, the 81-row hold-set and the 12/12 possessive hold-set all stand.

Worth carrying forward as a probe-design note rather than a code change: **a write-guarded probe cannot exercise any path that reads back what it just wrote.** The AI backfill lane is one of those, so this harness measures retrieval and ranking faithfully and cannot say anything about AI-estimated lines.

evawlve:
## Eval gate: RED, and the RED is the cache, not the fix

Ran the full golden set against a real build of this branch on the box — isolated tree `/tmp/gate160`, `next start` on :3100, identity verified (`readlink /proc/$PID/cwd` = `/tmp/gate160`, no `EADDRINUSE` in the server log) before trusting a single number. `FoodMapping` snapshotted first: 3,245 rows, 199 `human-triage`.

```
❌ 4 real failures, 🟡 15 known issues (expected)
```

**All four real failures are the four cases I added in this PR**, and every one of them returned its *pre-fix* answer:

```
❌ [n-mq-43] "trader joes cold brew"          -> Cold Pressed Juice, Red
❌ [n-mq-44] "trader joes turkey burgers"     -> Trader Joe's Turkey Meatballs
❌ [n-mq-45] "trader joes unexpected cheddar" -> Unexpected cheddar chicken sausage
❌ [n-mq-46] "trader joes scandinavian swim." -> Pane Guttiau Paper Thin Scandinavian Crackers
```

Those are cache hits. All four keys still hold the batch-01 rows, all `validatedBy = 'ai'`:

```
brew cold joe trader             | Cold Pressed Juice, Red                       | ai
burger joe trader turkey         | Trader Joe's Turkey Meatballs                 | ai
cheddar joe trader unexpected    | Unexpected cheddar chicken sausage            | ai
joe scandinavian swimmer trader  | Pane Guttiau Paper Thin Scandinavian Crackers | ai
```

**A filter fix cannot surface through a cache hit.** The route reads `FoodMapping` before it reaches retrieval, so the eval measures what the cache holds, not what the code now computes. This is already written down as the lesson on `n-mq-41` and in the investigation playbook, and I still walked into it: my probes ran `skipCache: true`, the eval does not.

The 15 known issues are the expected set with **no drift** and no new entries.

### What I changed rather than papering over

I did **not** mark these four `knownIssue`. A failing case makes the post-deploy cache refresh non-optional; a `knownIssue` would let it be forgotten and the fix would silently never take effect for real users. Each note now carries its own eviction statement.

### The other thing the eval caught — in my own work

`n-mq-47` was reported as **"known issue NOW PASSING"**. It was not fixed. It passed because `expectName: ["chicken breast"]` and both numeric bands are satisfied by the *wrong-brand* record — Mary's Chicken has the right food and the right macros, and the assertion cannot see a brand it does not name. **That is the exact defect the case exists to pin, and my assertion could not express it.**

Added `forbidName: [["mary's"]]`. Safe in the sense `n-mq-40` warns about: `mary's` is not a substring of `Trader Joe's Boneless Skinless Chicken Breast`, so the case becomes passable the moment the brand is right rather than unpassable forever.

This is the general hazard worth naming: **a wrong-brand pick with right macros passes every numeric band we have.** It is why this class survived every gate in the campaign, and it is what the correctness screen is being built to catch.

### Sequencing

Per the rule recorded on `n-mq-39` — *do not evict a key until the cold answer is known to be better* — these four qualify: the cold answer was measured before any eviction, 4/4, on both trees. The refresh is a 4-row production write and is being held for explicit approval rather than done inline.

Gate server on :3100 has been shut down (killed by port, identity confirmed first, live service on :3000 verified healthy after).

evawlve:
### Qualifying "1/17 → 6/17 correct"

Wiring the correctness screen into the campaign runner surfaced that my headline number measures less than the word "correct" implies. Restating it precisely.

The group-A probe scores a case with `correctName`, a **regex over the record name**. The hand audit that produced the 81 labels scores identity *and* the nutrition panel *and* the serving anchor. Those are different questions, and on two rows they give opposite answers:

| key | probe | audit |
|---|---|---|
| `kirkland muffin signature` | `<< ALREADY CORRECT` | **BAD** — `off_0826088442101` declares 263.2 kcal with **0 g protein and 0 g fat** (Atwater 1.25, impossible) and `servingGrams = 1.9`; `corruptReason` empty |
| `chili crunch joe onion trader` | correct | **BAD** — identity right, record corrupt and unmarked: the product's own ~1 tbsp label stored as a per-100g panel, `servingGrams = 1.0` |

So the honest form of the claim is **"1/17 → 6/17 by record identity"**, and the single case that was already passing before this PR is one whose *name* is right and whose *numbers* are not. The pool evidence, the zero regressions and the 81-row and 12-row hold-sets are unaffected — those are all before/after comparisons of the same measurement. But "correct" was the wrong word for the axis I measured.

I did check the four flips on the other axes by hand, and they hold: 240 g / 10 kcal for a black cold brew, 113 g / 180 kcal for a turkey burger, 28 g / 110 kcal for an ounce of cheddar, 28 g / 100 kcal for an ounce of gummy candy.

**This is the argument for the screen existing, made accidentally.** An identity probe and a panel audit fail on different rows: the probe blessed a muffin with impossible macros; the screen (#161) flags exactly that row via D6. Neither instrument alone is a correctness gate. That is also why the screen keeps Tier D *and* Tier L rather than collapsing to whichever looks stronger in isolation.

Verified by running the shipped screen CLI from #161 against real batch-01 data — reconstructed `added.txt` from the before/after psv pair (81 rows, matching the recorded count), real Postgres row pull, Tier D only so it is deterministic:

```
screened 81 rows | policy=balanced | tiers=D only | EVICT 19 · REVIEW 8 · KEEP 54
TIER D (any rule)              BAD 19/23  GOOD 0/48  SUSPECT 8/10
TIER D (EVICT only, balanced)  BAD 18/23  GOOD 0/48  SUSPECT 1/10
```

Matches the matrix #161 pins, through the real CLI on real data rather than a fixture.

evawlve:
### Scope correction: this PR fixes the candidate side only

A systematic audit of every tokenizing split in `src/lib/mapping/` and `src/lib/search/` found that **this fix has a mirror it does not close**, and the title overclaims. Recording it here rather than quietly expanding the PR.

**`hasApostrophe` is computed from the candidate string.** An apostrophe-free candidate therefore never gets the fold — while an apostrophe-bearing *query* still derives must-have token `amy` from `amy's` (`deriveMustHaveTokens`, `filter-candidates.ts:2944`, unfolded), which cannot match `Amys`. Measured through the real `filterCandidatesByTokens` on **this branch**, query-glyph × candidate-glyph:

```
Amy's | Pizza Cheese     mustHave  none=["amys","cheese"]   apos=["amy","cheese"]
      cand:   none    '    ’    ‘    ʼ    ´
  q=none       Y    Y    Y    Y    Y    Y      <- this PR working
  q='          .    Y    Y    Y    Y    Y      <- mirror fork, still live
```

Exposure: **2,170 brands exist in both spellings; 65,160 OFF records carry the bare spelling of a brand that also appears possessively** — comparable to the 55,858 possessive-spelled records this PR does fix. So this closes roughly half the class.

I knew about this gap when I chose the additive construction — the additive fold is what makes the change provably admit-only, and folding the query side *by substitution* is exactly what drops the `wendy's` pairs. Closing the mirror needs the same additive treatment applied to the query token, which is a second mechanism and a second set of measurements. It is not a reason to hold this one, but the PR title should be read as **"possessive-brand *records* survive the required-token filter"**.

### Second gap, in code this PR introduces

`foldApostrophes` folds `['’`]` — three glyphs. `canonicalizeCacheKey` folds `['‘’ʼ`´]` — six. The corpus carries the missing three: `trader joe‘s`, `hershey‘s`, `reese‘s`, `campbell‘s`, `kellogg´s`, `uncle ben‘s` — **112 possessive OFF rows across 70 brands**.

```
Trader Joe's | Scandinavian Swimmers
      cand:   none    '    ’    ‘    ʼ    ´
  q=none       Y    Y    Y    .    .    .     <- this PR covers ' and ’ only
```

Widening a fold character class is admit-only and cannot change any measurement already taken on this branch (those cases use `'` and `’`). I have deliberately **not** pushed it while the adversarial review is still reading this code — changing the function under review mid-review is how a review stops meaning anything. It goes in as a follow-up commit here once the review lands, together with making the class one shared exported constant instead of four copies that have already drifted.

### What the audit cleared

`filter-candidates.ts:2760` — the site I flagged as most suspicious in the brief — is **safe**, with a reason: `brandWords` has one consumer, a spare list at `:2773` whose only `return false` is exact equality against 30 punctuation-free literals, and both sides carry the same `len > 3` filter. Folding there would cause *rejections*, not fix any: it changes which `UNRELATED_INDICATORS` token survives for 8 of 13,051 possessive brands, all in the permissive direction (`grill'd`, `margarita's amigos`).

Also cleared with reasons: `brand-detector.ts:454/470` (index-aligned; candidate text never reaches that module), `cache-key.ts`/`cache-key-core.ts` (#149 holds — all spellings converge given the same `normalizedName`), `count-label.ts`, `build-fatsecret-result.ts` (fails closed), `learned-synonyms.ts` (inert), `validated-mapping-helpers.ts`, `query-token-coverage.ts`, and `filter-candidates.ts:498/547/593` (dead code, zero callers).

### One premise in my own brief was wrong

I told the audit the brand lexicon emits an apostrophe-free form like `"trader joes"`. It does not — `brand-detector.ts` echoes whatever the user typed and matches on exact spelling. Of 1,864 entries, 190 are straight, 31 curly, 0 exotic, and **165 straight-spelled brands have neither a curly nor a bare twin**, so `amys` → null, `mcdonald’s` → null, `wendy’s` → null. That kills `targetBrand` for brand-targeted retrieval, the decisive gate, the save gate and the cache-key brand prefix simultaneously.

The corollary matters for everything downstream: there are **three** query spellings in circulation — `joe's`, `joe s`, `joes` — not two.

The `joe s` one is now #162.

evawlve:
## Adversarial review landed. Two of my claims do not survive it.

Full independent review, ~83 tool calls, run against both trees on the box. It could not break the patch — *"I did not find a defect in the patch"* — but it broke two of the five measurement claims, and one of those undermines the precision of every diff-based number I posted. Correcting the record.

### 1. There is an 8.3% per-row noise floor, and I did not measure it

Two runs of the **same base tree** on the **same** 265 seeds returned **22 rows (8.3%) with a different winning record**. Flips cross sources (`fdc_171098`↔`fs_11126016`, OFF↔FS) and include THREW/not-THREW variation.

The cause is my own probe design, and it is the *second* consequence of the flaw I already retracted a finding over: the write guard blocks `aiNormalizeCache` upserts (150–175 per run), so uncached seeds re-query the LLM on every run. 235 of 265 seeds are uncached — and so are **7 of the 12** in my tripwire hold-set.

So: **"1 of 48 GOOD rows moved" and "0 of 12 changed" are at or below the noise floor and are not established.** I reported them without error bars against noise I never measured. The pool 1→10 flips are far larger than the floor and have an obvious mechanism, so those stand; the fine-grained safety numbers do not.

### 2. The change is not admit-only at the pipeline level

My "by construction" argument is about the required-token predicate, and the review confirms it holds there by reading every line. But the *pipeline* is not admit-only — the rerank pool **shrank** on 8 of 265 rows:

```
nyes cream sandwiches llc ice cream sandwich   pool 10 -> 1   winner CHANGED
ben jerrys netflix chillld ice cream           pool  5 -> 1   winner CHANGED
land ofrost oven roasted turkey breast         pool 10 -> 4   winner CHANGED
mckenzies lentils                              pool 10 -> 5   winner CHANGED
```

Three mechanisms downstream of the filter: `filtered.slice(0, 10)`; the all-drop guards at `map-ingredient-with-fallback.ts:1550/1590`, which stop protecting a bad-but-only candidate as soon as a second one exists; and relaxed-recovery suppression at 1598-1616, where a query that used to get the *wide relaxed* pool (which bypasses steps 3b–3e entirely) now gets the narrow strict one. I noted slice displacement once in passing for the tzatziki row; the commit message's framing does not, and should have.

### 3. The real hold-set — 265 rows, not 12 — and it found 2 regressions

The review built what the plan actually asked for: all **277** live `FoodMapping` rows with an apostrophe in `brandName`, deduped to 265, cold on both trees. On the 243 stable across two base runs: **6 improvements, 14 lateral, 2 regressions.**

```
mrs cubbisons caesar salad restaurant style croutons
    BASE   Caesar Salad Croutons [Mrs. Cubbison's]        pool=7
    BRANCH Caesar salad croutons [Reese]                  pool=5
gardettos special request roasted garlic rye chips
    BASE   ... [Gardetto's]        30g 130kcal  pool=2
    BRANCH Garlic rye chips []     28g 149kcal  pool=5
```

Both lose a correct brand, and both are now *cacheable* — the exact hazard this PR concedes in the abstract and had no measured instance of. And the sharpest illustration of why 12 hand-picked queries were not a hold-set: **`dennys 2 buttermilk pancakes` moves from the exact Denny's record (170 kcal) to an unbranded one (86 kcal, −49%)** — one word away from `dennys pancakes`, which my hold-set reports as unchanged. (That row is nondeterministic in base, so it is not attributed to the patch.)

Blast radius I never quantified: folding creates ≥1 new token on **86,606 of 93,467** apostrophe-bearing live OFF records — 9.1% of the corpus.

### 4. Tzatziki is lateral on the axes I checked, and not on the one I didn't

| | fs_69274 (before) | off_0002000496937 (after) |
|---|---|---|
| kcal/100g | 117 | 125 |
| protein/100g | **3.33** | **1.79** |
| sugars / fiber / sodium | 3.33 / 0 / 217 mg | **null / null / null** |

Identity, brand, serving kcal and serving size are lateral — precisely the set I checked. The panel is not: protein halves and three fields go from present to absent. Not corrupt (`corruptReason` NULL, Atwater self-consistent), just thinner. For a yogurt dip 3.33 g/100g is the more plausible figure.

This is the same axis-blindness I wrote up in the playbook two hours ago, committed again by me in the same session. Checking four axes and calling it "lateral" is the identity-probe mistake wearing a different hat.

### Fixed in `bd0e0d3`

**Two negative controls were vacuous.** Both "folding must not become substring matching" cases carry no apostrophe, so `hasApostrophe` is false and the folded branch is never entered — there was no test that a *folded* name fails to create a false match, the one hazard the block claims to cover. Added two that do enter it. Also moved `trader joes chicken breast` out of the additive-survivor block: that block's contract is "passes before **and** after", and this case fails before, which made the comment untrue.

Worth recording: `git stash push <file>` is a **no-op once the fix is committed**, and it silently reported a green "without fix" run. Use `git checkout origin/master -- <file>`. Now 5 fail / 10 pass against the real master file, 15 pass with the fix.

### Confirmed clean

Trees differ only in the patched file — verified structurally *and behaviourally* (`survivors=0 => UNPATCHED` / `survivors=1 => PATCHED`), and `node_modules` is hardlink-identical (same inode, 12 links). **No false-match hazard**: across all 93,467 apostrophe-bearing records folding creates 6,783 new tokens, all possessive collapse or French elisions (`dolive`, `lhuile`, `capn`, `smore`) — nothing spurious. **Folding the backtick is correct**: all 21 live backtick records use it as a possessive. **Latency is noise**: +2.4% (681.8 → 698.4 µs) against an ~18 ms p50 budget. `FilterResult.reason` is read nowhere; `removedCount` only in a log line.

### Two things the review caught that are mine to carry

- **`flip.ts` never resets `__LAST_RR`** (unlike `holdset.ts:54`), so cases that skip `simpleRerank` inherit the previous case's pool telemetry — 2 contaminated rows per file. None of n-mq-43..47 are affected, so the golden-set pool figures are clean.
- **`isAbstention` does not cover the AI-backfill lane.** An AI-estimated record carries a real `foodId` and non-zero confidence, and its `displayName` is the AI's rendering of the query — so for 4 of my 5 golden cases, whose `expectName` is a substring of their own query text, the gram floor is the only backstop. Only n-mq-45's `cheddar cheese` is not a substring of its query.

### Where this leaves the PR

The reviewer's verdict: *"on the population the PR targets, the change is 6 improvements to 2 regressions with 14 lateral moves — directionally good, and the mechanism is sound. I did not find a defect in the patch. I did find that the evidence is less precise than it reads."*

I agree with that, and I am **not merging on my own judgement**. The code is sound and the direction is positive, but I published safety numbers that were noise, and there are now two measured brand-loss regressions that will be cached. Diego's call.

PR #181, opened 2026-07-27
fix(eval): close the remaining class-B fail-open holes (flywheel/warm, parity sweep, stress, detect-corrupt-*)

evawlve:
Part of handoff §5a Phase 1 (task 1a) — **do not merge until adversarial review completes.**

## What changed

Playbook §11 class B (absence encoded as a PASS), the holes PR #177 did not close. Every fix follows the `detect-panel-scale-divided.ts` fail-closed pattern: refusals counted and printed, nonzero exit on failure, one deliberate fail-injection test per instrument (offline, mocked at the module boundary), each with positive controls.

### 1. flywheel-sweep.ts + warm-cache.ts (one unit) — the headline
**The exact mechanism defeating PR #177's `evalExitCode` fix:** `run-eval.ts` writes its results file *before* exiting — including on the zero-case path where `evalExitCode([])` makes it exit 2 — and `runEvalGate()` never read `proc.status` when a new results file existed. It re-derived pass/fail from `data.results ?? []`, and an empty array yields zero unexpected failures = **PASS**, exit 0 into systemd. run-eval's own fail-closed red was being converted to green by its most important consumer.

- New pure module `scripts/eval/flywheel-verdict.ts` (`judgeEvalGate`, `sweepVerdict`), wired into the sweep. Exit code is judged FIRST: exit 2, signal kill, spawn failure, missing/unreadable receipt, zero-case file, and a receipt that *contradicts* the exit code (stale file) are all instrument failures — never a pass, never a plain "gate red".
- Sweep exit codes, propagated to systemd via `process.exitCode`: **0** clean · **1** gate ran and FAILED · **2** instrument failure / measured nothing (warm total failure, zero seeds, zero results included). The verdict + reasons also land in the flywheel JSON (`exit` block) and stderr.
- Post-hoc gate made explicit (playbook §2): a gate-failure exit now *names* the warm writes that already happened ("NOT rolled back") — unchanged ordering, no longer silent.
- `ALLOW_FAIL` visibility: suppressed real failures and **stale** allow-fail entries are printed in console + markdown. Semantics unchanged (the nightly with default `n-mq-10` still goes red on `n-supp-23`, as it already did since #167 — now with reasons).
- `warm-cache.ts` standalone: `warmExitCode` exits 2 on zero seeds / zero results / total request failure; NaN `--concurrency` can no longer build a zero-worker pool (`Array.from({length: NaN})` = no workers, "ok 0 · errors 0" looked clean).
- `--stuck-keys-only` / `--seg-replay-only` / `--coverage-only` exit 2 when the single report they exist to produce failed (in-sweep these steps stay fail-soft/report-only by documented design).

### 2. cache-parity-sweep.ts
Overwrites cache rows as an intended side effect, yet printed `"changedRecord": 0` and exited 0 on **total HTTP failure**. Now distinguishes, exactly as specified:
- `0 changed record(s) over N rows verified reachable` → exit 0;
- partial transport failure → exit 1 ("those cache rows were verified by NOTHING this run");
- **nothing reachable / empty `--before`** → exit 2 + explicit `NOTHING REACHABLE … "0 changed records" is VOID` error.
Summary gains `verifiedReachable` + `exit`; diff logic extracted to pure `summarizeParity`/`parityExitCode`; `main()` previously had **no `.catch`** — crashes now exit 2 deterministically.

### 3. stress-latency.ts
Zero requests executed (`--n 0`, NaN repeat) exited 0 → now exit 2. New unambiguous hard-fail: EVERY should-hit search returning empty (dead index answering HTTP 200 over nothing — the header names empty-rate as a thing this script exists to catch, but it never gated on it). Existing error/null-nutrition fails preserved. Now importable (`require.main` guard — it used to self-execute on import, which is why warm-cache had to mirror its corpus lists).

### 4. detect-corrupt-nutrition.ts / detect-corrupt-panel.ts
A scan that saw ZERO rows (wrong DB, filter typo) wrote `Flagged 0 rows (of 0 scanned)` + a report file and exited 0. Now: exit 2 and **no report written**, so a failed scan can never fake an empty population for `mark-corrupt-off.ts`. Both scans take an injectable `ScanDb` (same shape as the reference detector's `RowStream`) and no longer construct PrismaClient at import. Report format unchanged.

### 5. Sweep of the rest of scripts/eval/*.ts — zero-input guards (empty in = green out)
`apply-repoints.ts` (0 repoints → "Plan: 0 updates" exit 0), `classify-drops.ts` + `triage-drops.ts` (0 rows → a report reading "no drops"), `probe-bare-serving.ts` (0 stdin queries), `measure-serving-divergence.ts` (0 fixture rows; all anchors errored → agreement measured against nothing). All now exit 2 with explicit errors.

### Audited, NOT changed (deliberate — listed so the cap is explicit)
- `serving-provenance.ts`: zero-row DB prints a zeros report + exit 0. One-off diagnostic; whether an empty `OffServing` table is instrument failure is ambiguous. Left as-is.
- Flywheel telemetry/stuck/seg-replay/coverage failures **inside a full sweep** remain fail-soft report-only (documented contract; seg-replay explicitly "must never change the sweep's exit code").
- `classify-drops.ts`/`triage-drops.ts` `attachIncumbents` catch-and-warn ("counts read as unknown") — visible degradation, left.
- `winner-diff.ts`/`winner-diff-screens.ts`: already hardened by #172/#177 (noiseGate, exitCode 1/2 paths); not re-audited beyond grep this pass.
- `correctness-screen.ts`: out of scope (parallel D8 agent owns it); `screenExitCode` already fail-closed from #177.
- `ops/systemd/flywheel-sweep.service` untouched (deploy-only box; `Type=oneshot` already surfaces nonzero exits).

## Numbers
- Baseline reproduced exactly: `npx jest scripts/eval/__tests__` **708 passed, ~13s**, offline.
- After: **743 passed** (+35 in `__tests__/fail-open-sweep.test.ts`); full suite **111 suites, 2,333 passed, 1 skipped**; `typecheck` clean; `lint:ci` 0 errors (448/700 warnings, pre-existing).
- Method rule honored: fixes were broken on purpose — reverting the exit-2 branch, the zero-case branch, and the parity total-failure branch each turned a named test red (1, 1, and 1 failures respectively), then restored.
- Wiring proven by grep: `judgeEvalGate`/`sweepVerdict`/`warmExitCode`/`parityExitCode`/`stressExitCode` all reached from their runners; no caller still re-derives pass from a results file.

## Discrepancies from the spec
- None material. One nuance worth review attention: the old flywheel *did* exit 1 on a red gate — the defeat was confined to the **exit-2 / zero-case / contradictory-receipt** shapes (and to error paths that produced no results file being indistinguishable from gate-red). The PR makes red-vs-void a first-class distinction rather than introducing exit-on-red.
- The three writing scripts (`warm-cache`, `flywheel-sweep`, `cache-parity-sweep`) and `stress-latency` were exercised ONLY through the offline jest suite with mocked I/O, per the task's safety constraints — none were run against production.

### DIFF
--- a/scripts/eval/__tests__/fail-open-sweep.test.ts
+++ b/scripts/eval/__tests__/fail-open-sweep.test.ts
@@ -29,7 +29,8 @@ import * as path from 'path';
 import {
     EvalRunEvidence, judgeEvalGate, sweepVerdict,
 } from '../flywheel-verdict';
-import { runWarm, warmExitCode, WarmResult } from '../warm-cache';
+import { runEvalGate } from '../flywheel-sweep';
+import { main as warmMain, runWarm, warmExitCode, WarmResult } from '../warm-cache';
 import { parityExitCode, summarizeParity } from '../cache-parity-sweep';
 import { Sample, stressExitCode } from '../stress-latency';
 import {
@@ -147,6 +148,54 @@ describe('judgeEvalGate: run-eval exit 2 can never be re-derived into a PASS', (
     });
 });
 
+describe('runEvalGate: the evidence ASSEMBLY judged against a real child, not a hand-built receipt', () => {
+    // The 35 judgeEvalGate tests above prove the judge; none of them prove
+    // that the spawn's exit code, the before/after eval-* file diff and the
+    // JSON parse actually REACH it. This block spawns a real child — a local
+    // node stub, no network, no DB — that reproduces run-eval's fail-closed
+    // shape: write a plausible results file FIRST, then exit 2.
+    const writeStub = (dir: string, exitCode: number): string => {
+        const stub = path.join(dir, `stub-run-eval-exit${exitCode}.js`);
+        fs.writeFileSync(stub, [
+            "const fs = require('fs');",
+            "const path = require('path');",
+            'const resultsDir = process.argv[2];',
+            "fs.writeFileSync(path.join(resultsDir, 'eval-stub-' + process.pid + '.json'),",
+            "    JSON.stringify({ results: [{ id: 'n-a-01', query: 'apple', detail: '', pass: true }] }));",
+            `process.exit(${exitCode});`,
+        ].join('\n'));
+        return stub;
+    };
+
+    it('a child that writes a plausible results file and THEN exits 2 comes out RED through the real assembly', () => {
+        const resultsDir = tmpDir();
+        const stub = writeStub(resultsDir, 2);
+        const gate = runEvalGate({
+            spawn: { cmd: process.execPath, args: [stub, resultsDir] },
+            resultsDir, allowFail: [], timeoutMs: 30000,
+        });
+        expect(gate.ran).toBe(false);
+        expect(gate.pass).toBe(false);
+        expect(gate.error).toContain('exited 2');
+        expect(gate.error).toContain('does not override');
+        // ...and the sweep-level consequence systemd sees: instrument failure.
+        expect(sweepVerdict(gate, { seedCount: 10, resultCount: 10, okCount: 10 }).code).toBe(2);
+    });
+
+    it('POSITIVE CONTROL — the same stub exiting 0 passes, so the assembly (not the stub) is what decides', () => {
+        const resultsDir = tmpDir();
+        const stub = writeStub(resultsDir, 0);
+        const gate = runEvalGate({
+            spawn: { cmd: process.execPath, args: [stub, resultsDir] },
+            resultsDir, allowFail: [], timeoutMs: 30000,
+        });
+        expect(gate.error).toBeUndefined();
+        expect(gate.ran).toBe(true);
+        expect(gate.pass).toBe(true);
+        expect(gate.casesRun).toBe(1);
+    });
+});
+
 describe('sweepVerdict: the sweep exit code systemd sees', () => {
     const goodGate = () => judgeEvalGate({
         status: 0, signal: null, resultsFile: '/x/eval.json',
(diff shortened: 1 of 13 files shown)


evawlve:
## Review follow-ups applied (commit 8793fad)

All three targeted items from the adversarial reviews, on this branch — no new PR since #181 already carries `fix/fail-open-instruments`.

### 1. IMPORTANT — `cache-parity-sweep.ts`: `parityExitCode` was blind to `unresolvableBefore`
The reviewer probe (3 rows whose before-export has `offBarcode`/`fdcId`/`fsId` all null) reproduced exactly as claimed: `summarizeParity` → `{total:3, sameRecord:0, errors:0, unresolvableBefore:3, verifiedReachable:3}` → exit 0 and a success line over rows that were diffed against **nothing** — the same absence-encoded-as-clean class this PR closes, inside the function it added.
- `errors + unresolvableBefore >= total` is now **exit 2 VOID** ("NOTHING DIFFED … ZERO identity comparisons ran").
- `verifiedReachable` now **excludes** unresolvable rows (`total - errors - unresolvableBefore`), so the printed denominator counts only rows an identity comparison actually ran on; the success line says so explicitly.
- Tests: the reviewer probe verbatim (asserts exit 2 + `verifiedReachable === 0`), a partial case (2 diffed + 1 unresolvable → denominator 2, exit 0), and a combined case (1 error + 1 unresolvable = whole run → exit 2 even though neither alone reaches total).

### 2. MINOR (both reviewers) — `warm-cache.ts`: `--dry` returned before the zero-seed guard
The guard now sits **above** the `--dry` fork so both paths share it, wired through the real `warmExitCode` zero-seed branch. `main()` is exported with injectable argv + seed assembly and **returns** its exit code (the `require.main` block applies it), so the wiring itself is under test: `--dry` over a zero-seed corpus returns 2, the wet path shares the guard, and a positive control (`--dry --limit 5`) lists exactly 5 seeds, returns 0, and provably never touches `fetch`.

**Discrepancy report (per spec):** the literal repro in the item — "`--dry` with an empty/bad `--seed` file prints 'Warm corpus: 0 names'" — does not reproduce on this branch: `--seed` **adds** to the standard corpus (built-in lists are unconditional, so seeds ≥ ~120 whenever the repo is intact), and a *nonexistent* `--seed` path throws in `assembleSeeds` → the existing catch exits 2. The zero-seed state needs the corpus assembly itself to come back empty (missing/partial checkout, or a future `assembleSeeds` change). The fail-open wiring was real regardless — the dry path bypassed the guard — and is now closed + regression-tested via the injectable assembler.

### 3. MINOR — `flywheel-sweep.ts`: `runEvalGate` evidence assembly now has a stub-child integration test
`runEvalGate` is exported with an injectable child + results dir (defaults byte-identical to production; `main()` calls it bare). The new test spawns a **real local node stub** that reproduces run-eval's fail-closed shape — writes a plausible `eval-*.json` results file, *then* exits 2 — and asserts the gate comes out RED through the real spawn/status/file-diff/JSON-parse path, plus `sweepVerdict` exit 2. Positive control: same stub exiting 0 passes with `casesRun: 1`. Fully offline.

### Numbers
- Suite: `npx jest scripts/eval/__tests__` **743 → 751** (+8), 13 suites, all green. `typecheck` clean, `lint:ci` 0 errors (448/700 warnings, unchanged class).
- Diff: 4 files, +235 / −27.
- Method (playbook §11): all three fixes were **mutation-checked** — each reverted on purpose (`verifiedReachable` denominator + VOID branch, `!dry &&` on the guard, `status: 0` forced in the assembly) and the 5 new red-path tests went red exactly, with the positive controls staying green; then restored and re-verified.

PR #182, opened 2026-07-27
chore(eval): commit cache-ops scripts; evict population guard fails CLOSED per-key

evawlve:

## What changed

The cache eviction/restore tooling for the 2026-07-27 audit existed only as untracked files in `/tmp/ra-clone` — one reboot away from destroying the only rollback path for a production cache eviction. This PR commits them onto current master (file CONTENT taken from `/tmp/ra-clone`, never its git state — that checkout is 3 PRs stale), and closes two fail-open holes (playbook §11 class B):

### `scripts/eval/_evict_rows.ts` — population guard now fails CLOSED
- **Was:** compared FoodMapping row COUNT only. Proven fail-open: the 04:35 flywheel-sweep updated **305 rows in place** while the count stayed **3,248**, and evict-set key `and ben jerry` was fully re-resolved to a DIFFERENT food — the guard still passed.
- **Now:** every evict-list key's live identity fields are content-checked against the snapshot row; on ANY mismatch the ENTIRE run refuses (exit 2) and prints every moved key with per-field diffs. Never skip-and-continue — the operator re-cuts the list. Both pre-existing refusals kept (key missing from snapshot; count moved). Keys-file and snapshot-file parsing also refuse on empty/truncated/missing/internally-inconsistent input instead of reading as "nothing to do".
- Identity fields compared: `source, foodName, brandName, offBarcode, fdcId, fsId, validatedBy`. **Spec discrepancy:** the task named "foodId / offBarcode / fdcId / validatedBy", but `FoodMapping` has no `foodId` column — its record pointer is the `offBarcode`/`fdcId`/`fsId` trio; `foodName`/`brandName` are included because they are the identity the screen's record card actually judged. Heartbeat fields (`usedCount`/`lastUsedAt`/`validatedAt`/`aiConfidence`) deliberately excluded, and rows OUTSIDE the evict list may move without refusal — otherwise the guard can never pass on a live cache.

### `scripts/eval/_restore_rows.ts` — skips counted, printed, distinct exit code
- `createMany({skipDuplicates:true})` kept — never clobber a re-resolved row is CORRECT. But the silent skip is the limit on reversibility, so now: every skipped key is printed (full list, not a sample), createMany race skips (created < submitted) are counted too, and **exit 3** (distinct from 0=clean, 2=refused) whenever skips > 0.

### Committed byte-identical from /tmp
`_snap_foodmapping.ts`, `_dump_cache_prompts.ts`, `_dump_prompts.ts` (playbook §10 bake-off recipe references it), `_recover_seeds.ts` (the §10 seed-recovery method — pins-only attribution). **Left out:** `_probe_eventlog.ts` — one-shot diagnostic ("does MappingEventLog.normalizedForm live in cache-key space?") whose settled answer is encoded in `_recover_seeds.ts`'s header; not referenced by any documented recipe. Nothing else from `/tmp/ra-clone` crossed over.

### Restructuring
Guards are pure exported functions; `PrismaClient` construction moved inside `main()` behind `require.main === module`, so the offline suite imports the modules with no DB and no mocks. CLI usage, flags, dry-run-by-default, and output shape unchanged.

## Tests

`scripts/eval/__tests__/cache-ops-guards.test.ts` — 27 tests, fail-injection pattern from `fail-closed.test.ts` (PR #177), fixtures under `__tests__/fixtures/cache-ops/`:
- (a) row updated in place under an UNCHANGED count → refuses, names the key and fields (the `and ben jerry` incident shape, plus source-flip and validatedBy-flip variants; one moved key refuses the whole run)
- (b) key missing from snapshot → refuses; moved count → refuses
- (c) empty / truncated / missing / count≠rows.length / zero-row snapshot → refuses
- (d) restore: every skipped key printed, exit 3 on skips (incl. race skips), positive controls throughout

**Mutation check (playbook §11):** disabled the identity comparison → 4 tests red; disabled the skip exit code → 2 tests red; reverted → green. The tests demonstrably watch the guards.

## Numbers

- `npx jest scripts/eval/__tests__`: **708 → 735** (13 suites), master baseline 708 reproduced before changes. (PR #181's 743 is on its own branch; the two PRs share no files.)
- `npm run typecheck`: clean. `npm run lint:ci`: 0 errors.
- Scripts were **never executed** — exercised only through the offline jest suite. No DB, no network, no production access.
- Also done outside this repo (plain file copies, no commit): `audit-evict-keys.json` (320 keys) + `audit-review-keys.json` (116 keys) → mobile `sync-docs/cache-snapshots/`, verified byte-identical with `cmp`.

## Discrepancies
- Spec's "foodId" identity field does not exist on `FoodMapping` (see above) — implemented as the record-pointer trio + name/brand, a strict superset.
- Handoff pins no count for `audit-review-keys.json`; it contains 116 keys (for the record: 104 UNSURE + 17 serving-axis = 121 ≠ 116; not investigated — out of scope, list content untouched).


evawlve:
## Follow-up: remaining IMPORTANT finding + two minors applied (commit c8b11f0)

### ⚠️ CLI BREAK — `_evict_rows.ts` no longer accepts a positional snapshot path

The command documented in `handoff_cache_audit_2026-07-27.md` §1(a) **will now refuse with exit 2**. Corrected command line:

```bash
cd /tmp/ra-prev && npx ts-node --project tsconfig.scripts.json --transpile-only \
  -r tsconfig-paths/register scripts/eval/_evict_rows.ts \
  /tmp/screen-llm/audit-evict-keys.json \
  --screen-snapshot /tmp/screen-llm/snapshots/FoodMapping-pre-evict.json --execute
```

(The `.claude/settings.local.json` permission rule from handoff §5a Phase 0 still matches — it is prefix-anchored on the script path, not on the argument shape.)

### 1. IMPORTANT — the two snapshot ROLES are now explicit

The identity guard is only meaningful against the **at-screen-time** snapshot (Role A, verdict anchor). Handoff §2b says to re-take a snapshot immediately before any `--execute` — that fresh file is the **restore anchor** (Role B) for `_restore_rows.ts`, and an operator who fed it to `_evict_rows` got a vacuously-green identity check (live trivially matches a seconds-old snapshot). Now:

- snapshot argument is the named flag `--screen-snapshot`; a bare second positional **refuses**, explaining both roles — it cannot be guessed at, because the failure mode is invisible at run time. Unknown flags and a duplicated `--screen-snapshot` also refuse (a typo'd role flag must not silently un-anchor the run).
- run start prints a banner: which file anchors the verdicts, its `takenAt` age (**loud STOP warning when the "screen" snapshot is < 15 min old** — that file is almost certainly Role B on the wrong flag), and the procedure step (2 of 5 dry run / 4 of 5 execute).
- the five-step two-file procedure (screen snapshot → guard; fresh snapshot → restore anchor) is documented in the headers of `_evict_rows.ts`, `_snap_foodmapping.ts`, `_restore_rows.ts`.
- the post-delete hint used to say `Restore with: _restore_rows.ts <screen-snapshot> <keys>` — i.e. it taught the role confusion in the *other* direction. It now points at the fresh Role-B file.
- `_snap_foodmapping.ts` stamps a role-neutral `label` + `roleNote` meta into the payload (a snapshot has no inherent role; the operator assigns one by use) and says so on stdout.
- fail-injection: the exact old handoff-§1a command shape is a pinned refusing test.

### 2. minor — `_recover_seeds.ts` zero-recovered is VOID, not success

`recovered 0 / N` now exits **4** (`RECOVER_VOID_EXIT`, distinct from 0=ok and 2=error) with an explicit message: an empty pin file silently un-pins every row downstream (§11 class B). Partial recovery stays exit 0 — positive control pinned at the real 1,868/3,248 shape. Prisma construction moved inside `main()` behind `require.main` so the offline suite imports it.

### 3. minor — `_dump_cache_prompts.ts` refuses a broken pins.tsv

Empty/whitespace pins file, any line that is not `key<TAB>phrase` (truncation mid-line, extra tabs, empty phrase), and well-formed pins that overlap **zero** row keys (wrong key space / wrong file) all refuse with expected-vs-observed (malformed count + line numbers + first examples), exit 2, **before any batch file is written**. Previously all three read as "0 pins, carry on" with exit 0.

### Numbers

- `npx jest scripts/eval/__tests__`: **743 → 770** (14 suites), baseline 743 reproduced on the untouched branch first.
- Mutation check (playbook §11, break-on-purpose → red → restore): positional-tolerant CLI restored → **2 red**; zero-recovered mapped to exit 0 → **2 red**; empty-pins fail-open → **2 red**; malformed-line silent skip → **6 red**; all restored → 770 green.
- `npm run typecheck` clean; `npm run lint:ci` 0 errors (448 pre-existing warnings, unchanged class).
- Nothing did not reproduce; no scripts were executed — offline jest only, no DB, no network to LAN hosts.

PR #184, opened 2026-07-27
feat(eval): OffFood snapshot script — per-row-restorable, fail-closed (Phase 1d)

evawlve:
Part of handoff §5a Phase 1 — **do not merge until adversarial review completes.**

## What this is

The `OffFood` snapshot instrument — the hard prerequisite and ONLY rollback path for the §4 divided-panel corpus repair (which multiplies numeric panel fields by `servingGrams/100` on ~3,173 rows; a wrong flag inflates a 1,060 g row by 10.6×). No snapshot mechanism existed: `grep -rn pg_dump` finds only a comment in `cache-parity-sweep.ts:9` telling a human to do it by hand.

**BUILD AND TEST ONLY. This script was NOT executed against any host or database this session.** The single real validation run happens separately, only after Diego names that authorization. Defaults are structured so that run needs zero code changes: `--host owner@192.168.1.133`, `--container mealspire-db`, `--out-dir /home/owner/snapshots` (all overridable by flag or `SNAPSHOT_*` env).

## Design

- **Runs from a dev machine, dump lands ON the DB host** (`ssh → docker exec psql COPY ... TO STDOUT | gzip -9 > /home/owner/snapshots/OffFood-<ts>.tsv.gz`), so it survives dev-machine reboots. Only counts/checksums/column lists cross ssh.
- **COPY text format, not `pg_dump -t`**, on documented reasoning (full rationale in the file header):
  1. per-row restore is the requirement (the repair may be partially rolled back) — side table + `UPDATE ... FROM` on `barcode`;
  2. COPY text = exactly one line per row, so **dump lines must equal a same-run `SELECT count(*)`** — a mechanically checkable invariant `pg_dump` does not offer;
  3. explicit column list = checkable coverage: columns discovered live from `information_schema`, refused unless a superset of pinned `REQUIRED_COLUMNS` (every column the §4 repair could write: `nutrientsPer100g`, `corruptReason` via the mark stage, `updatedAt`, plus the whole rest of the row).
- **Fail-closed at every step (playbook §11 class B)**: transport failure, docker/psql/COPY failure, count mismatch, count-after drift (table moved), empty table, empty/short dump, gzip corruption, column-shape drift, unparseable count (**empty stdout is NOT zero**), and pre-existing output file (timestamped name + refuse-overwrite) — all exit nonzero with explicit expected-vs-observed. The dump is written to `.partial` and only renamed after ALL verifications pass: a file bearing the final name IS a verified snapshot.
- **The `embedding vector(384)` question**: default is the safe complete choice (ALL columns). `--measure` is a read-only sizing mode (rows, relation sizes, embedding column bytes; dumps nothing) to settle whether excluding it matters; `--exclude-embedding` exists for after that measurement. **UNSETTLED: live dump size / embedding share — needs the authorized run.**
- Restore semantics documented in the header, printed after every successful run, and stored in a `.meta.json` manifest (row count, sha256, column list, restore recipe) next to the dump.

## Numbers

- Offline gate `npx jest scripts/eval/__tests__`: **708 (master baseline) → 774**, all green — 66 new tests, all offline, single mocked transport boundary, no network/DB/ssh.
- Fail-injection coverage per the task spec: transport failure, pg COPY failure, count mismatch, empty file, existing-file refusal — plus count-after drift, empty table, missing/unsafe columns, verify-label absence, mv/manifest failure, and a positive control (a suite that only refuses proves nothing).
- §11 method rule: three deliberate breakages (empty-count→0 fail-open, mismatch check deleted, overwrite check forced ABSENT) each turned the suite red (3, 1, 2 failures) before revert.
- `npm run typecheck` clean; `npm run lint:ci` 0 errors, 449/700 warnings.

## Discrepancies

- `npm run lint` (`node scripts/lint.js`) exits 1 on **clean origin/master** in this environment (rushstack ESLint patch: "calling module was not recognized") — pre-existing, unrelated to this change; CI's `build` check uses `lint:ci`, which is green.


evawlve:
Applied all three review items in `e0a58c3` (offline only — nothing was run against any host or database).

**1 (important) — Typesense re-sync is now step 6 of the rollback recipe.** The recipe (success output + `.meta.json` manifest, one source: `restoreInstructions()`) previously ended at `DROP TABLE "OffFood_restore"`; since the search path serves nutrients straight off the Typesense hit (`gather-candidates.ts` maps `hit.nutrientsPer100g` into the candidate), a Postgres-only rollback leaves search billing the un-rolled-back numbers indefinitely. The recipe now carries the exact command — full rebuild `scripts/sync-typesense.ts` on the DB host — placed **before** the side-table drop, plus the same warning in the header docs. On the mode question: **full rebuild is correct even for a bounded row-set restore**, because (a) the repo's only bounded updater (`updateTypesenseDocumentsByFilter`) PATCHes one *identical* partial doc into every filter match — right for the `hasCountServing` flag backfill it was built for, wrong for per-row panel values; (b) live `off_foods` doc ids can predate id=barcode keying (documented in `typesense-client.ts`), so a hand-rolled per-row upsert can *duplicate* docs instead of replacing them; (c) a restore can change `corruptReason`/`duplicateOfBarcode`, which changes which docs should *exist* at all — only the rebuild re-applies those exclusions (`purge-corrupt-typesense.ts` already names it as the recovery path).

**2 (minor) — `.partial` cleanup on failure.** Post-dump failures now issue a best-effort `rm -f` of the *current run's* `.partial` — exact filename, never a glob; pre-dump failures skip it (no partial can exist, no pointless ssh); a failing or throwing cleanup never masks the nonzero refusal and points at the new HOUSEKEEPING header section, which documents the exact `ls`/`rm` pattern for orphans (only `*.tsv.gz.partial`; never the verified `*.tsv.gz` / `*.meta.json`).

**3 (minor) — bash is now guaranteed, in code.** `sshTransport` wraps every step as one remote-command string `bash -c '<script>'` via the new exported `sshArgs()`. Docstring was indeed wrong before: sshd hands the string to the remote *login* shell, and `set -euo pipefail` needs bash/zsh. A test also pins against the other bad shape — passing `bash`/`-c`/`script` as separate argv entries, which ssh space-joins unquoted so the login shell word-splits the script before bash sees it.

Numbers: fail-injection suite 66 → 77 tests; `scripts/eval/__tests__` 774 → 785 passed (13 suites); typecheck clean; lint:ci 0 errors. Per playbook §11, each fix was broken on purpose after commit and its new tests went red on exactly the broken item (3 / 4 / 2 tests), then restored to green. Nothing failed to reproduce; no behavior changes outside `snapshot-off-food.ts` and its test file.

PR #185, opened 2026-07-27
feat(eval): gated divided-panel corpus repair script (handoff §5a-2a) — BUILD ONLY, do not merge without review

evawlve:
## What this is

The data-op counterpart to the PR #179 detector, for **the biggest live user-facing defect on the board**: the OFF chain import divides the whole per-100g panel by `servingGrams/100`, so 3,173 rows / 824 families / 55 brands bill 5–10x low (a Jersey Mike's Giant and a Regular both bill 173 kcal against true ~1,831/~937; 203 FoodMapping rows point into this import, 72 materially distorted).

**BUILD AND TEST ONLY.** Nothing here has touched any database. The dry run against the real corpus happens later under Diego's named authorization; `--execute` is Phase 3.

> **UNSETTLED:** 3,173 is a **FLOOR**, not the class size — singleton defective rows are invisible to the family test (a one-size product never forms a multi-size family). Closing the gap needs the FatSecret/FDC cross-source join — a separate queued task, not this script.

## Design (handoff §4 + §5a-2a + corrections table; all constraints inherited, none re-tuned)

- **Detection is imported, not re-implemented** — family key, slope band, r², barcode restriction, member guard and the plausibility guard all come from `detect-panel-scale-divided.ts`. The detector's `prismaStream` is now exported (the only change to that file) so both scripts scan the *identical* query — a re-implemented stream is playbook §11 class A, the known failure.
- **Repair:** every finite numeric field of the raw stored panel × S/100. `buildPlan` asserts the raw scaling agrees with the detector's `repairPanel` on every canonical field and refuses the whole plan on disagreement (class-A tripwire).
- **The plausibility guard gates every single write**, recomputed from the LIVE row at write time; refusals are counted and listed, never dropped. Precedent: it caught `Taco Bell, Crunchwrap Supreme` on the detector's first live run (the 260g row is already a correct density — "repairing" it would invent a 108.1 g/100g macro sum). That exact shape is a test here.
- **`Asda | Thai sticky rice` (50571723466684) is hand-excluded** via a named, commented constant — enforced twice: dropped into a visible `exclusions` list at plan build, and *any* plan carrying it refuses validation outright (corrections table: the detector emits it as a repair).
- **Risk tiers:** tier 1 = families with ≥3 distinct sizes (strong evidence); tier 2 = 2-size families, where a 2-point log-log fit is perfect *by construction* (weak evidence). The plan separates them; `--execute --tier 1` lets Diego approve tier 1 alone.
- **Gating built into the script** (`decideMark()` has no rule for this class, so the gated mark/purge/replay method is not executable here):
  - `--execute` **refuses to even parse** without `--plan <approved dry-run artifact>` and `--snapshot-manifest <verified snapshot-off-food.ts .meta.json>` (shape + verification fields validated: table, positive rowCount, sha256, required repair-writable columns, FINAL non-`.partial` filename, restore recipe — and the manifest must be **newer than the plan**: order is plan → approve → snapshot → execute).
  - Each row is **re-read immediately before its own write** and refused per-row if any recorded before-value moved (name, brand, servingGrams, the whole raw panel, corruptReason, duplicateOfBarcode) — the §11 population-guard lesson applied per row; the count-only guard was proven fail-open on 2026-07-27.
  - The written value is **recomputed from the live row**, never copied from the plan; a tampered `rawAfter` refuses as plan-drift.
  - Plan validation refuses entries **below the 150g member guard** — load-bearing, because the plausibility guard is one-sided (ceilings) and a small-S entry would *shrink* a panel under every ceiling — plus duplicate barcodes, detector `ScanReport`s masquerading as plans, wrong arm/version, and empty plans.
  - An execute that applies **zero** rows exits 2 — "0 changed records" with exit 0 is the exact fail-open shape of `cache-parity-sweep.ts`.
- **Dry run emits** per-family stats (slope, r², sizes), per-row before→after for every numeric field (full detail in the plan JSON), guard refusals with reasons, risk tiers, per-brand totals, and a deterministic hand-audit sample (first N by `(familyKey, barcode)`) — the artifact Diego approves in one look.
- **After writes**, a mandatory-next-steps block prints: FULL Typesense rebuild (`scripts/sync-typesense.ts` — the search path serves nutrients straight off the Typesense hit, `gather-candidates.ts:160`; incremental patching would be wrong per-row and can duplicate docs) and the FoodMapping implications (203 rows point into this import; **no eviction needed** — identity-only cache, panels hydrate from OffFood per request; verify with a live JM Giant/Regular probe). Unblocks warm batches 07–10.

## Testing (offline only — no DB, no network, per the standing safety constraints)

- **52 new fail-injection tests** (`scripts/eval/__tests__/repair-panel-scale-divided.test.ts`), each refusal block with a positive control: guard refusal counting, the Crunchwrap shape at plan time AND reaching for the database at write time, Asda exclusion (constant + plan build + validation), snapshot-manifest refusals (absent/malformed/unverified/stale/`.partial`/missing columns — 15 mutations), per-row moved-value refusals (panel/serving/corruptReason/deleted), plan-drift, tier filtering, CLI gate refusals, class-A agreement between `scaleStoredPanel` and the detector's `repairPanel`.
- Full `scripts/eval/__tests__` suite: **17 suites, 979 passed** (was 927 before this PR; the detector's own 28 tests untouched and green with the export change).
- `tsc --noEmit` clean. (`npm run lint` fails at ESLint startup in this sandbox on untouched master too — environmental; the `build` check here is the real lint gate.)

## What was NOT done, deliberately

- No script was run against any database, in any mode. The authorized read-only dry run is:
  ```
  cd /tmp/ra-prev && git fetch origin && git checkout feat/repair-panel-scale-divided && \
  npx ts-node --project tsconfig.scripts.json --transpile-only -r tsconfig-paths/register \
    scripts/eval/repair-panel-scale-divided.ts --out /tmp/panel-scale-repair-plan.json
  ```
- `--execute` additionally needs the approved plan + a fresh verified OffFood snapshot manifest, and is blocked on Diego by design.

### DIFF
--- a/scripts/eval/__tests__/repair-panel-scale-divided.test.ts
+++ b/scripts/eval/__tests__/repair-panel-scale-divided.test.ts
@@ -23,6 +23,7 @@ import {
 import {
     applyPlan,
     buildPlan,
+    checkPlanSha256,
     checkSnapshotCoversPlan,
     executeExitCode,
     EXCLUDED_BARCODES,
@@ -33,9 +34,11 @@ import {
     mandatoryNextSteps,
     parseRepairArgs,
     scaleStoredPanel,
+    sha256Hex,
     tierOf,
     validateManifestText,
     validatePlanText,
+    verifySnapshotOnHost,
     DEFAULT_SAMPLE_SIZE,
     PLAN_KIND,
     UNSETTLED_FLOOR,
@@ -43,8 +46,9 @@ import {
     type PlanEntry,
     type RepairDb,
     type RepairPlan,
+    type SnapshotManifest,
 } from '../repair-panel-scale-divided';
-import { REQUIRED_COLUMNS } from '../snapshot-off-food';
+import { REQUIRED_COLUMNS, type ExecResult, type Transport } from '../snapshot-off-food';
 
 // ===========================================================================
 // Fixtures — same generators as panel-scale-divided.test.ts, so the fixture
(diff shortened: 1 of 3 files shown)


evawlve:
Follow-up commit `438d262` applies the four review items from both adversarial reviews:

1. **`--tier` is now REQUIRED in execute mode** (was: silent default to `all`, the widest write scope). The refusal names the tiers, their corpus row counts (~2,093 tier-1 / ~1,080 tier-2 / ~3,173 all) and points at the plan's `tierTotals` for the exact counts. Dry-run behavior unchanged.
2. **Tier labels are recomputed, not trusted**: `validatePlanText` re-derives `tierOf()` from each entry's own `familyDistinctServings` and refuses a mismatch — a hand-relabelled tier-2 entry can no longer be selected by `--tier 1`. Entries missing `familyDistinctServings` refuse too.
3. **`--sample` refuses in execute mode** (consistent with the existing `--out` refusal) — `--execute --sample 10` must not silently apply the full tier.
4. **(a)** The dry run prints the plan file's sha256; `--execute` recomputes it from the exact bytes of `--plan` and refuses a mismatch against the new optional `--plan-sha256` flag (omitted ⇒ the computed hash is printed prominently for eyeball comparison, never skipped silently). **(b)** New `verifySnapshotOnHost()`: before any write, the dump the manifest names is re-verified **on the DB host** (exists + sha256 matches) over `snapshot-off-food.ts`'s own `bash -c`-wrapped ssh `Transport` — imported, not re-implemented (§11 class A). Manifests without `sshHost` now refuse. Execute-only; tests mock the Transport boundary.

New execute-mode CLI shape:
```
repair-panel-scale-divided.ts --execute --plan <approved-plan.json> \
  --snapshot-manifest <OffFood-*.meta.json> --tier 1|2|all \
  [--plan-sha256 <64-hex hash the dry run printed>]
```

**Verification**: suite 979 → 998 tests (17 suites), all green; typecheck + lint:ci clean. Every gate was mutation-checked — each fix broken on purpose, its test went red (6/6: tier default restored, recompute disabled, sample allowed, sha mismatch ignored, MISSING dump accepted, host-hash mismatch ignored), then restored byte-identical.

PR #193, opened 2026-07-30
feat(doc-check): fail-closed doc-claims registry + nightly checker

evawlve:
Docs were the last artifact in this project with no executable gate. Two tidy-up
passes had already failed, because tidying does not make rot detectable — a
command that fails does.

## What this adds

- **`scripts/doc-check/doc-check-core.ts`** — all verdict logic as pure functions
  with no I/O, so the fail-closed tests can see it. The runner is deliberately
  thin (the `run-eval.ts` lesson: a tested predicate wired into an untested
  verdict is an untested verdict).
- **`scripts/doc-check/__tests__/doc-check-core.test.ts`** — 22 tests. Every
  fallible input is forced to fail and every block carries a positive control.
- **`scripts/doc-check/claims.json`** — 40 claims across backend / mobile / box,
  each with the claim as its owner doc states it, the ONE doc that owns the fact,
  the command that re-derives it, the expected value, and the date measured.
- **`ops/systemd/doc-check.{timer,service}`** — nightly 04:50 on the box.
- **`npm run doc-check`**, and **`npm run eval:golden`** (see below).

## Fail-closed decisions worth reviewing

- A command that did **not** run FAILS its claim. The box being unreachable is
  exactly when its state is least verified, so it is never a skip.
- `parseRegistry` rejects an empty `expect.value` at parse time. `equals ""`
  would let a dead ssh or a missing file read as a pass; emptiness must be
  asserted as `count 0` over a `wc -l`, where the command's shape proves it ran.
- An uncompilable claim regex is a FAIL, not a pass.
- `checkerExitCode([])` is **2**, not 0 — a filter that matched nothing is not a
  green run. Same rule the golden eval already applies to zero cases.
- `--where backend,box` prints every excluded claim id. The box has no copy of
  the mobile repo, so its nightly covers 25 of 40; a silent cap would read as
  "covered everything" when it didn't.

## Verified

- `npm run doc-check` → **41/41 hold** from the Mac (all contexts).
- On the box, `--where backend,box` → 23/23, and a real `systemctl --user start
  doc-check.service` returned `Result=success`, `ExecMainStatus=0`.
- `npx jest scripts/doc-check` → 22/22 · `scripts/eval` → 17 suites / 1,004.
- `npm run typecheck` → 0 · `npm run lint:ci` → 0 errors.

## Also here, found while writing the claims

- **`eval:golden`** — `CLAUDE.md` named `scripts/eval/run-eval.ts` but nothing
  invoked it, and `npm run eval` is a different harness over `eval/gold.*.csv`.
- **`failure-classes.ts`** cited `662d609` for the FatSecret persist-race fix.
  That is PR #146 coverage-trend, and `git merge-base --is-ancestor 4fab5f1
  662d609` exits 1 — it is not even an ancestor of the real fix (PR #159). It
  also claimed all ten `markSaveRejected` sites are covered; nine are, and the
  tenth reports a write that threw rather than a gate that rejected.
- **`classify-drops.ts`** told readers to join event rows on
  `canonicalizeCacheKey`, which the registry's own caveat refutes and which
  produced 8 impossible "cold" events. Now names `deriveFunnelCacheKey`.
- Comments citing two handoffs repointed to `sync-docs/archive/` in the mobile
  repo, where those handoffs moved.

No runtime code is touched; the box needs no redeploy for this.


evawlve:
Added `32111b9` — registers `docs-reader-builds` (`where: mobile`).

Context: the mobile repo gained `npm run docs`, an HTML reader over `sync-docs/log/` + `sync-docs/reports/` (`scripts/docs-reader/build.mjs`, mobile `de6a95a`). It parses write-off frontmatter and `## Needs Diego` sections, so it breaks silently if that format drifts — this claim makes that a checker failure instead. Owner doc is the new `## Reading them` section in the mobile repo's `sync-docs/log/README.md`.

`npm run doc-check` (Mac, all contexts) → 42/42 claims hold.

PR #214, opened 2026-08-02
Lane-composed rerank window + the four FatSecret serving defects it exposed

evawlve:
> **Draft on purpose.** The change does what it claims, and the winner-gate refutes it as shippable. Both facts are the deliverable. Do not merge until the serving defect below is fixed.

## The defect is real, and larger than the handoff recorded

`filtered.slice(0, 10)` took a **prefix** of gather order. `gatherCandidates` concatenates its lanes (FDC → OFF → FatSecret) rather than interleaving, so whenever the earlier lanes filled the window, every later lane was **deleted before `simpleRerank` ran** — not out-ranked, never seen.

Measured cold over 249 already-asked lines (`winner-diff snapshot`, which forces production flags, so the FatSecret lane is live exactly as on the box):

```
249/249  queries truncate
207      (83.1%) lose at least one ENTIRE lane
fatsecret    gathered for 249, reached the reranker on 42  (16.9%)
fdc          0.0% starved
openfoodfacts 3.2% starved
```

It is not a `grilled chicken breast` quirk. It is the default behaviour.

## The fix

`buildRerankPool()` round-robins across lanes in **first-appearance order**, preserving within-lane order, with the **same budget** (`RERANK_POOL_LIMIT` === the old literal `10`).

- **Not a source preference** — no lane is ranked above another; lane order is inherited from gather.
- **Not an admission relaxation** — `|window|` is unchanged. It is a re-composition, and re-composition is non-monotone in *both* directions.
- Leaf-safe module (no imports, structural `{ source }` typing) specifically so the eval harness can import the real function rather than transcribe it.

13 tests. All five mutations kill their named test — and **one mutation survived the first draft** (untagged candidates becoming one lane each); the test that closes it was written from the mutation, not from the code.

## What the gate said

Noise floor **0/20 and 0/250 on both trees**, variant `gate-backstop` on both sides, serving stage on.

**Cold (20 seeds): 15 SAME, 5 WINNER-CHANGED, 0 WINNER→NOWINNER.**

`n-mq-22`'s item moves into band — `21.4P` (OFF, 95 kcal/100g) → `29.6P` (FatSecret, 195 kcal/100g), against the case's `[25, 38]`. Every tripwire held: `chipotle chicken burrito bowl`, `impossible burger patty`, `mac and cheese`, `shrimp taco`, `kirkland almonds`. Class drift: BASE 2, BRANCH 0.

**Regression (250 already-asked): 93 WINNER-CHANGED (37%) — and this is the refutation.** 73/250 (29%) change what the user is **billed**:

| query | before | after | tier |
|---|---|---|---|
| `1 tbsp peanut butter` | 16.0 g / 95 kcal | **7.5 g / 44 kcal** | `volume_unit` → `fs_volume_density` |
| `2 tbsp peanut butter` | 32.0 g / 190 kcal | **15.0 g / 88 kcal** | same |
| `0.5 cup peanut butter` | 125.0 g / 743 kcal | **60.0 g / 353 kcal** | same |
| `1 cup greek yogurt` | 250.0 g / 310 kcal | **120.0 g / 110 kcal** | same |
| `almonds` | 28.0 g / 160 kcal | **1.2 g / 7 kcal** | `bare_plural_serving` → `fs_label_count` |
| `1 courgette` | 196.0 g / 41 kcal | **28.4 g / 5 kcal** | `seed_count_default` → `fs_default_serving` |

## Root cause — and it is not this change

On `1 tbsp peanut butter` the winner moved from an OFF peanut butter (594 kcal/100g) to a FatSecret peanut butter (588) — **the same food, 1% apart**. The record did not get worse. The **serving tier** changed.

`fs_volume_density` in `build-fatsecret-result.ts` falls back to a flat `isLiquid ? 1.0 : 0.5` g/ml when no volume-quantified serving exists. Peanut butter is ~1.07 g/ml, so `1 tbsp = 14.79 ml × 0.5 = 7.4 g` against a true ~16 g — a **~50% under-bill on every non-liquid volume request that lands on a FatSecret record without a volume serving**. `almonds` bills **one almond** because `fs_label_count` noun-matches a `1 almond (1.2g)` serving on a bare plural query.

These tiers are latent today **precisely because of the starvation** — FatSecret wins 16.9% of the time. Any change that raises that rate detonates them. This one raises it to ~53%.

This is the PR #173 lesson at scale, and exactly what the playbook says to look for: *winner-diff gates which record wins, not what the user is billed.*

## Ordering

1. Fix `build-fatsecret-result.ts`: the volume-density fallback, and the bare-plural `fs_label_count` path.
2. Re-run this gate.
3. Then ship this.

Shipping as-is would be **worse than the defect it fixes** — and worse still under warming, where those grams become cached `FoodMapping` rows.

## Also here (both pre-existing, both needed to gate this honestly)

- **`winner-gate.sh` never passed `--variant`**, so `resolveVariant()` defaulted to `baseline` — the pre-#168 caller, which no tree has contained since 2026-07-26. Every run this driver has ever done modelled a caller neither side had, and a Step-4 change is invisible on the ~10.6% of lines where the gate fires. Now resolved from the tree's own caller hash; aborts rather than guessing.
- **`WinnerInfo.inTop10` is positional, not window membership.** Under lane composition those are different facts; `inRerankWindow` is the real one.

## Receipts

Caller re-pinned `16bec3f864088b83`. `verify --from-events --limit 60`: MATCH 60, MISMATCH-UNEXPLAINED 0, CONF-UNEXPLAINED 0. `npx tsc --noEmit` clean. Mapping suite: 3 pre-existing 5s-timeout flakes — clean master fails **4** of the same suite under identical conditions.

### DIFF
--- a/scripts/eval/winner-diff-screens.ts
+++ b/scripts/eval/winner-diff-screens.ts
@@ -71,6 +71,8 @@ export interface WinnerInfo {
     selectionReason: string;
     /** index in `filtered` (GATHER order) — >= 10 means it was outside slice(0,10) */
     indexInFiltered: number;
+    /** Positional: index in `filtered` < 10. NOT window membership — the rerank
+     *  window is lane-composed, so use `inRerankWindow` for "reached the reranker". */
     inTop10: boolean;
     inRerankWindow: boolean;
 }
--- a/scripts/eval/winner-diff.ts
+++ b/scripts/eval/winner-diff.ts
@@ -461,6 +461,20 @@ const KNOWN_CALLERS: Record<string, SelectionVariant> = {
     // gate — the one playbook section 5a requires before any admission or ranking
     // change — could not be run without `--allow-drift`. It was not run.
     '9e5634f738b23ba7': 'gate-backstop',
+    // CURRENT — the shape section 9 transcribes. Delta from 9e5634f7 is the rerank
+    // WINDOW: `filtered.slice(0, 10)` (a prefix over gather order) became
+    // `buildRerankPool(filtered, RERANK_POOL_LIMIT)` (per-lane composition), and the
+    // count-label extension now walks the remainder by difference instead of
+    // `filtered.slice(10)`.
+    //
+    // Still 'gate-backstop', and the variant taxonomy is deliberately NOT extended:
+    // `SelectionVariant` models what the confidence GATE does relative to Step 4
+    // (pre-empt vs backstop), and that is untouched here. Adding a variant per
+    // window shape would mean section 9 has to reproduce both, which is only worth
+    // doing when one tree must produce both sides of an A/B. This change is gated
+    // the other way — two trees, each transcribing its own caller — so the base
+    // tree's own pin (9e5634f7) covers the base side.
+    '16bec3f864088b83': 'gate-backstop',
 };
 /**
  * Re-pinned 2026-08-01 alongside the `copiedHelperSource()` CRLF fix above.
@@ -497,7 +511,7 @@ const PINNED_HELPERS_HASH = '7535716631ca65c8';
  * (that shape is genuinely known, and naming it is more useful than "unrecognised"),
  * but it can no longer claim the replay mirrors it.
  */
-const TRANSCRIBED_CALLER = '9e5634f738b23ba7';
+const TRANSCRIBED_CALLER = '16bec3f864088b83';
 
 interface DriftResult {
     caller: string;
@@ -594,6 +608,13 @@ const { AI_NUTRITION_MAX_PER_BATCH, AI_NUTRITION_HYDRATION_MAX_PER_BATCH } = req
 
 // Everything below is IMPORTED FROM THE REAL MODULES, never reimplemented. A probe
 // that reimplements the caller's argument list is a different function.
+//
+// `rerank-pool` is imported for the same reason, and it was written leaf-safe so
+// that it could be: no imports of its own, structural `{ source }` typing rather
+// than `UnifiedCandidate`. Requiring it here pulls in nothing from `config.ts`
+// (which snapshots flags and warms ONNX — playbook section 4).
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { buildRerankPool, rerankPoolRemainder, RERANK_POOL_LIMIT } = require('../../src/lib/mapping/rerank-pool');
 const { confidenceGate } = gatherMod;
 const { filterCandidatesByTokens, hasCoreTokenMismatch, hasNullOrInvalidMacros, detectGrainCookingContext } = filterMod;
 const { simpleRerank, toRerankCandidate, stripPrepModifiers } = rerankMod;
@@ -1307,13 +1328,32 @@ function replaySelection(e: SnapshotEntry, debug: boolean, variant: SelectionVar
         // caller 1657
         const countedNoun = countedPieceNoun(parsed);
 
-        // caller 1660-1668: NON-MONOTONE #1 — slice over GATHER order, not score
-        // order; the count-label extension pulls from filtered.slice(10) to a hard
-        // cap of 13. In BOTH variants this is now the ONLY way a candidate with
-        // indexInFiltered >= 10 can reach the reranker.
-        const candidatesForRerank: UnifiedCandidate[] = filtered.slice(0, 10);
+        // NON-MONOTONE #1 — the rerank window.
+        //
+        // The caller no longer takes a PREFIX of `filtered`. It composes the window
+        // per retrieval lane (`buildRerankPool`), because a prefix over gather order
+        // deleted whole lanes: measured cold 2026-08-01, 207 of 249 already-asked
+        // lines lost at least one entire lane, and the FatSecret lane reached the
+        // reranker on 42 of 249.
+        //
+        // IMPORTED, NOT TRANSCRIBED. `rerank-pool.ts` is leaf-safe by construction
+        // (no imports at all, structural `{ source }` typing rather than
+        // `UnifiedCandidate`), specifically so this harness can call the real
+        // function. Every hand-copy in this file is a drift liability — section 2
+        // exists because of them — so where the caller's logic can be imported
+        // instead, it is.
+        //
+        // The window SIZE is unchanged (RERANK_POOL_LIMIT === the old literal 10),
+        // so this is a re-composition, not an admission change. It is still
+        // non-monotone: a candidate that reaches the reranker under a prefix can be
+        // evicted by lane composition, which is exactly what this harness is for.
+        //
+        // The count-label extension now walks the REMAINDER (by difference), not
+        // `filtered.slice(10)` — the window is no longer a prefix, so the old form
+        // would re-offer candidates already inside it. Hard cap of 13 unchanged.
+        const candidatesForRerank: UnifiedCandidate[] = buildRerankPool(filtered, RERANK_POOL_LIMIT);
         if (countedNoun) {
-            for (const c of filtered.slice(10)) {
+            for (const c of rerankPoolRemainder(filtered, candidatesForRerank)) {
                 if (candidatesForRerank.length >= 13) break;
                 if (copy_candidateHasCountLabel(c, countedNoun)) candidatesForRerank.push(c);
             }
@@ -1637,6 +1677,12 @@ function mkWinner(
         confidence: clamped,
         selectionReason,
         indexInFiltered: idx,
+        // POSITIONAL ONLY. Since the rerank window is composed per lane rather
+        // than taken as a prefix, "in the first 10 of `filtered`" and "reached
+        // the reranker" are different facts. `inRerankWindow` below is the one
+        // that answers the second, and it is derived from the actual window ids.
+        // Kept because it still describes gather position, which is what the
+        // starvation analysis is about; do not read it as window membership.
         inTop10: idx >= 0 && idx < 10,
         inRerankWindow: rerankRan && windowIds.includes(w.id),
     };
--- a/scripts/eval/winner-gate.sh
+++ b/scripts/eval/winner-gate.sh
@@ -135,6 +135,36 @@ fi
 
 RUN='npx ts-node --project tsconfig.scripts.json --transpile-only -r tsconfig-paths/register scripts/eval/winner-diff.ts'
 
+# ------------------------------------------------------- blind spot A, part 2
+# THE VARIANT MUST BE THE ONE THE TREES ACTUALLY CONTAIN.
+#
+# `resolveVariant()` defaults to 'baseline', and this driver never passed
+# --variant, so every run it has ever done replayed `baseline` -- the PRE-#168
+# caller, where confidenceGate PRE-EMPTS Step 4 and simpleRerank never runs. No
+# tree has contained that shape since 2026-07-26 (`fb211474`). On the ~10.6% of
+# lines where the gate fires, both sides were therefore modelling a caller
+# neither tree has, and any change living inside Step 4 is INVISIBLE on exactly
+# those rows -- understating both its benefit and its risk.
+#
+# winner-diff says this out loud (`announceVariantFit`). The driver piped it to
+# `tail -6` and it scrolled past. A warning nobody can see is not a warning,
+# which is the same lesson as the drift banner it sits next to.
+#
+# Derived from the tree, never assumed: `hashes` resolves this tree's caller hash
+# through KNOWN_CALLERS. If the tree is UNRECOGNISED there is no honest variant
+# to pick and we stop -- that is the drift guard's job and it must not be routed
+# around here.
+TREE_VARIANT="$(eval "$RUN" hashes 2>/dev/null | sed -n 's/^this tree is: \([a-z-]*\) .*/\1/p' | head -1)"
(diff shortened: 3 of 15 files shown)


evawlve:
## Gate re-run with the volume-density fix (`3499698`)

Same harness, same two snapshots, noise floor **0/20 and 0/250 on both trees**, variant `gate-backstop`, `--with-serving`.

**Winners are identical** — 93 WINNER-CHANGED in the regression population, unchanged. Expected: the serving fix cannot move a winner. That is the control.

**The volume cluster is gone.** Movers whose new billed value came from `fs_volume_density`: **6 → 2**. Every case named in the PR body as a defect has dropped off the list entirely:

| query | before the fix | after |
|---|---|---|
| `1 tbsp peanut butter` | 16.0 g → **7.5 g** | no longer moves |
| `2 tbsp peanut butter` | 32.0 g → **15.0 g** | no longer moves |
| `0.5 cup peanut butter` | 125.0 g → **60.0 g** | no longer moves |
| `1 cup greek yogurt` | 250.0 g → **120.0 g** | no longer moves |
| `1 cup white rice` | 186.0 g → **120.0 g** | no longer moves |

Regression `GRAMS-CHANGED` 50 → 46, `SERVING-SAME` 151.

**Still draft — a second, distinct tier is unfixed.** The remaining movers are dominated by tiers this commit did not touch:

```
new-value tier, regression movers:   fs_default_serving 16   fs_label_count 6   fs_volume_density 2
```

`fs_label_count` is the bare-plural bug named in the PR body and deliberately not fixed here: bare `almonds` still bills **1.2 g** (one almond), `1 courgette` still bills 28.4 g against 196 g. That is a different mechanism — `inferDiscreteUnit` noun-matching a `1 almond (1.2g)` serving on a query that means a portion — and it wants its own change and its own gate run.

**Worth reading the remaining list carefully rather than treating every mover as a regression.** Several are improvements the old path was getting wrong:

- `coffee` 240 g / **91 kcal** → 177 g / **2 kcal** — black coffee is ~2 kcal; the base was wrong.
- `a bowl of cereal` **1.0 g** / 4 kcal → 33.0 g / 124 kcal — the base's 1 g is not a bowl of anything.
- `a tbsp of peanut butter` 32 g → 16 g — 16 g *is* a tablespoon; the base was billing the 2-tbsp label serving.

So the honest read is: one of the two serving blockers is closed and confirmed closed by the gate, the other is scoped and open, and the leftover set is a mix of real regressions and corrections. Do not merge until `fs_label_count` is dealt with and this is re-run.

evawlve:
## Gate run 3 — the bare-request serving path (`d026de3`, `2de5bf9`)

Noise floor **0/20 and 0/250 on both trees**. Winners unchanged at **93 WINNER-CHANGED / 156 SAME** across all three runs — the control holds, as it must: a serving fix cannot move a winner.

### `almonds` is fixed, and the first attempt at it was not

| | base | branch |
|---|---|---|
| gate run 2 (`3499698`) | 28.0g `bare_plural_serving` | **1.2g** `fs_label_count` |
| gate run 3 (`d026de3`, suppression only) | 28.0g `bare_plural_serving` | **1.2g** `fs_default_serving` |
| gate run 4 (`2de5bf9`, + band) | 28.0g `bare_plural_serving` | **28g** `bare_category_default` |

Suppressing the per-piece branches changed the tier and not the number, because FatSecret's own `defaultServingId` for `fs_37040` **is** the 1.2g "1 almond" row. Read off the box:

```
35301  1 almond                  1.2g   <- declared default
44764  1 oz                     28.35g
35300  1 oz (23 whole kernels)  28.35g
59771  100 g                     100g
35299  1 cup whole                143g
```

Every suppression drains into the default-serving branch, and that branch had no band. The fix rejects an out-of-band serving on a bare request and **does not substitute** — the prisma include has no `orderBy`, so "take the next in-band serving" yields 28.35g or 143g depending on DB row order. Falling through to the bare-query guard is deterministic: `almonds` → 28g, `coca cola` → 355g, both matching what `buildOffResult` already bills.

### Two live defects it fixes that were not the target

Both are **base-tree** rows, i.e. they are wrong today, independent of the lane-composition change:

```
black pepper    0.1g fs_default_serving  ->  2.5g bare_category_default
salt            0.4g fs_default_serving  ->  2.5g bare_category_default
```

### What the existing tests caught that I had wrong

The macro-only tier needed the same guard or the fix would have been inert again — inverting the 1-almond serving's 7 kcal against a 578 kcal/100g panel is 1.2g by another route. My first version applied the full 3–400g label band there and the panel-inversion tests killed it: a brewed coffee legitimately inverts to ~477g, and a 400g ceiling sent it to the flat 100g default. Only the **floor** transfers to that tier; those grams are arithmetic on the record's own panel (validated 98.1% within 5%), not a label.

One existing expectation changed deliberately: bare `diet coke` billed its `100 g` placeholder and now bills 355g via the guard. Its actual subject — never the 1g shape, always 0 kcal — is unchanged and still asserted.

### Still open, still blocking this PR

Same-class under-bills the band does **not** reach, because it is bare-only and these lines carry digits:

```
1 handful almonds          -> 1.2g    (still one almond)
1 sleeve saltine crackers  -> 3g
1 courgette                -> 28.35g against OFF's 196g   (FS has no seed count table)
peanut butter              -> 16g against the 32g category default
```

The last one is structural: `fs_default_serving` aliases to a CAP-only guard tier, so the guard can lower grams and **never raise them**. An under-bill on that tier is invisible to it by construction.

evawlve:
## Gate run 5 (`b084375`) — dose rule works; the `orderBy` has a regression

Noise floor 0/20 and 0/250 both trees. Winners still 93/156 — control holds.

**Dose-count reconciliation: confirmed.** `peanut butter` is off the mover list — 32g base, 32g branch, tier `bare_dose_count_reconciled`, description `2 tbsp (~32g)`, scaled from `fs_3384`'s own 16g tablespoon. No other row moved onto that tier, matching the prediction that only the nut-butter category can reach it.

**The `orderBy` is reproducible but ARBITRARY, and that is not good enough.** Three rows moved, two better and one clearly worse:

| query | before | after |
|---|---|---|
| `5 cherry tomatoes` | 745.0g | **85.0g** `5 x 1 cherry (17g each)` ✅ |
| `1 courgette` | 28.4g | **124.0g** (vs OFF's 196g) ✅ |
| `chicken` | 85.0g / 201kcal | **7.0g / 17kcal** ❌ |

`chicken` now bills `1 thin slice (approx 2" x 1-1/2" x 1/8") (7g)` off `fs_1623`.

Honest framing: the 7g row was **always reachable** — `usableServings[0]` was unordered, so this was a coin flip on Postgres row order, and the change made it deterministic rather than new. But "deterministic and wrong" is not shippable, and sorting by `servingId` asserts nothing about which serving answers the question.

Note the bare band does **not** catch it: `usableBareLabelServing()`'s floor is 3g, and 7 clears it. The rule that *would* catch it is `BARE_MIN_PIECE_SERVING_GRAMS` (20g), which today applies only to the `fs_label_count` path, not to the default-serving branch — the same asymmetry that let `almonds` through two commits ago, one tier over.

**Recommendation:** do not ship the `orderBy` alone. Either
1. extend the 20g piece floor to the default-serving branch when the serving description names a piece/slice — small, mirrors the rule already in this file, and would take `chicken` off the 7g row; or
2. revert the `orderBy` and land it together with the "prefer the record's own serving nearest the category default" re-pick, which is the real fix (measured: 23 of 40 candidate rows, 544 of 909 events, have such a serving on the record).

I have not applied either. #214 stays draft.

evawlve:
## Gate run 7 — both blocking regressions closed; recommending this comes out of draft

Commits since run 5: `9cc2a87` (declared-default coin flip) and `388cacd` (ambiguous-unit convergence).

**Gate:** noise floor **0/20 and 0/250 on both trees**. Winners **93 WINNER-CHANGED / 156 SAME** — unchanged across runs 5, 6 and 7, which is the control a serving fix must not move.

### `chicken` — the `orderBy` was not the cause

My previous comment offered two fixes and both were wrong, because I had the mechanism wrong. `defaultServingId` resolves for **23,837 of 23,837** FatSecret foods, so `?? usableServings[0]` fires in exactly one situation: the declared default *is* the shared `100 g` panel row that `isPer100gPanelServing` deliberately demotes out of `usableServings`. That is **569 foods**, and for every one of them the billed grams were decided by include order. Adding `orderBy` made that reproducible without making it right — and reproducible-and-wrong is worse than flaky, because it stops being noticed.

**278 of the 569 answer the question themselves** with an explicit `1 serving (Xg)` row, and 3,422 of the 3,565 such rows corpus-wide sit inside the existing 3–400 g band. Preferring it is the record's own statement of what a serving is; position is not evidence of anything.

It matters more than it looks because **no lexicon rescue exists for this class**: `getBareQueryDefault` returns nothing for chicken, beef, steak, salmon, pork, turkey, rice, pasta, bread, egg, ham or sausage. Falling through would have billed the flat 100 g.

Gate: `chicken` 113.4 g / 110 kcal → **85 g / 201 kcal**.

### The two digit-line regressions run 6 exposed

`buildOffResult()` and `buildFdcResult()` both resolve `handful` / `sleeve` / `bowl` / `knob` through `getOrCreateAmbiguousServing()`. `buildFatSecretResult()` **never called it** — the fourth instance this session of a rule owned by one module and adopted by some cascades but not others.

| line | base | run 6 | run 7 |
|---|---|---|---|
| `1 handful almonds` | 30 g | 1.2 g | **28 g** ✅ |
| `1 sleeve saltine crackers` | 113 g / 490 kcal | 3 g / 13 kcal | **113 g / 484 kcal** ✅ |

**Banding the declared default was the obvious fix and is wrong.** This builder is terminal — it always bills something — so a rejection lands on the flat per-100 g tier, and the bare-query guard cannot rescue these because its digit gate excludes them by construction. 100 g for a handful of almonds is not better than 1.2 g, only wrong in the other direction. The request needs an answer, not a veto.

It also fixed three base defects that were not targeted: `1 ramekin of hummus` 15 g → 150 g, `3 rashers of bacon` 15 g → 90 g, `3 strips bacon` 15 g → 45 g.

Cost: ambiguous units are **5,789 of 54,086 live events (10.7%)**, and 3,890 of those are cache hits. Same cost OFF and FDC already pay.

### A gate limitation worth recording

Of the 93 serving changes, **27 touch AI on at least one side and are NOT attributable to this branch** — `cooked quinoa` has the *same winner and same tier* in both trees and still moves 50 g → 185 g, because `fdc_size_estimate` calls an LLM. The noise floor measures **winner** diffs only, so it reports 0/250 while this variance is live. Only **66 of the 93** are deterministic. Anyone reading a `--with-serving` mover list should split on `aiTouched` before calling anything a regression.

### Knowingly shipping, stated rather than buried

`1 cup white rice` 186 g `fdc_label_volume` → 120 g `fs_volume_density`. This is **not new**: `1 cup brown rice` already billed 120 g in the base tree on the OFF density path, so 0.5 g/ml for rice is a pre-existing defect and this branch widens it to white rice by taking away its FDC label. Fixing it is a change to the density **numbers**, which `volume-density.ts`'s own header requires be a separate PR with its own gate. Not bundled here on purpose.

### Recommendation

Out of draft. `npx tsc --noEmit` exits 0 and `lint:ci` has 0 errors locally. Also worth correcting on the record: **#209 and #211 are already deployed** — the box runs `NC8Ouof85RBlInIgC_7yf` (2026-08-01 17:18), not the `lu9XX…` the docs claim. And `n-mq-22` is not an independent blocker: it fails on `protein100` 21.4 vs [25,38], which is scale-free and therefore a pick failure, and this PR's own gate records it moving to 29.6 P — in band.

evawlve:
## Correction after deploy: `n-mq-22` is still RED, and my prediction above was wrong

Merged and deployed (box `NC8Ouof85RBlInIgC_7yf` → `syHuRD63kgNpyzd9_AJ_9`, service active, `/api/ok` 200). The eval still reports:

```
❌ [n-mq-22] protein100=21.4 outside [25, 38] (mapped: "Grilled Chicken Breast")
   off_0850001478458 / openfoodfacts / 126g
```

Byte-identical to the pre-deploy run. **The comment above claimed this PR turns the eval green. It does not.**

**The code IS live** — `skinless chicken breast` now bills `fs_label_count` 236g, a tier the previous build could not emit. The reason `n-mq-22` did not move is that it never reaches the new code:

```
rawLine                | foodId            | servingTier        | cacheHit | funnelStage
grilled chicken breast | off_0850001478458 | bare_label_serving | early    | cache_hit
```

`cacheHit='early'` — retrieval never runs, so `buildRerankPool()` is never called. The cached row is `FoodMapping` key **`breast chicken grilled`** → `off_0850001478458`. The key is **token-sorted**, which is why querying `normalizedForm` for `'grilled chicken breast'` returns 0 rows and reads as "not cached" — I made exactly that mistake before finding it by barcode.

**The generalisable part, which is a limitation of the gate and not of this PR:** `winner-diff` replays a **frozen gather snapshot**, so it bypasses the cache entirely. A gate result is therefore a claim about **cold retrieval only**. It does not predict live behaviour for any line that already has a `FoodMapping` row — and 2,855 keys do. Every "the gate says this case moves" claim in this PR carries that caveat, including the 21.4P → 29.6P one.

Nothing here changes the serving-defect work, which is measured on grams and unaffected by which layer supplies the record. But the eval will stay red until the cached row is evicted, and eviction is a DB write I have not made.

PR #236, opened 2026-08-04
DO NOT MERGE (refuted by its own gate) — fix(rerank): run macro-consensus outlier demotion on generic queries

evawlve:
The one mechanism for "this row disagrees with its identically-named siblings" — `RANK_BRAND_CONSENSUS` in `simpleRerank()` — was gated on a detected `targetBrand`, so it never ran for the generic queries that need it.

`apple` resolves to a 23.9 kcal/100g record against a 52.8 sibling median (n=53); `unsweetened almond milk` to 60.0 against 16.5 (n=32). Both sit inside the corrupt-panel detector's direction bands and are rejected by its rescale clause — it is a panel-**scale** detector, so a plainly wrong panel that is not a serving-scale artifact is invisible to it.

## Population (measured 2026-08-04)

**29,394 unmarked live `OffFood` rows** sit ≥1.6× or ≤0.6× their identically-named sibling median in a group of ≥4 — 2.7% of 1,085,525 kcal-bearing rows over 27,650 groups. **97.9% are structurally invisible** to the panel-scale detector. Existing detectors already reach ~31% (13,397 marked). That is ~200× the two golden cases this was filed under.

## Approach

The branded path is byte-unchanged. Generic queries group siblings by `normalizeNameKey()` — the **same** key the corrupt-panel detector and dedupe already use. Two competing definitions were in play (raw `lower(name)` vs this one); picking the detector's keeps one owner. All existing safety rails are untouched: ≥3 nutrition-bearing siblings, majority agreement near the median, median floors. `simpleRerank()`'s own comment already argued those conditions — not the `targetBrand` gate — are the real safety rail.

## Also here

- **`ORIGINAL_SCORE`'s one-sided clamp documented as UNRESOLVED.** Flooring it at 0 was tried and reverted: it removes a penalty from badly-matching OFF rows, and OFF rows winning when they should not is this pipeline's dominant defect (95.2% of cross-source contests already put an OFF row at index 0). That is *reasoning, not measurement* — and a test written to pin the current behaviour proved **inert under mutation**, so the honest state is "unmeasured, so unchanged", recorded in the comment rather than in a fake test.
- **`n-mq-27` / `n-mq-47` promoted out of `knownIssue`** — 4/4 stable cold probes each plus a full cold run, against a 20/348 noise floor. **`n-cook-06` deliberately NOT promoted**: 5 probes gave 4 passes and one fail (`off_8801052053783` Konjac Cooked Rice oats, kcal100 105 vs band [45,95]).
- doc-check claim guarding the paused `flywheel-sweep.timer`, marked TEMPORARY with its own deletion instruction.

## Verification

- `npm run test:ci` → 155 suites / 3208 pass · `typecheck` 0 · `lint:ci` 0 errors
- **Mutation control executed, not just described**: restoring `if (targetBrand && ...)` kills 2 of the 3 new tests. The third pins the `normalizeNameKey` grouping.

## ⚠️ NOT GATED YET

**Predicted cold movement: `n-prod-01` and `n-mod-02` close (12 → 10 real failures); knownIssue 12 → 10 from the promotions.**

The gate must be a local baseline-vs-change pair on **one** host — the box serves `TqkY7MBZjJjmnKuqDjeoF`, which does not contain this change, so a cold run against it would measure the deployed pipeline and say nothing. Accept if the actual diff ⊆ predicted ∪ the recorded 20-case noise set.


evawlve:
## Gate run: REFUTED. Do not merge as-is.

Local baseline-vs-change cold pair, one host, both arms built and served locally with `FATSECRET_RETRIEVAL_ENABLED=1` (verified live at the consumer: `white rice` -> `fs_4501`, `source: fatsecret`). Both arms ran the SAME `golden-set.json`, so the two `knownIssue` promotions are constant across the pair and the diff is purely the code change.

- arm A (baseline `22b67ed`) build `TbpKTGHPSp2ahnAPwoLpl` — **13 real failures**
- arm B (this branch) build `aojHh0FLYX_uzPlqeKeF0` — **11 real failures**

**Prediction recorded on the commit: `n-prod-01` and `n-mod-02` close. Both are still failing.** The only two cases that changed state were `n-mq-41` and `n-serv-39`, neither predicted, and 2 changed cases sits well inside the recorded cold noise floor of 20/348. Under the predict-then-bisect rule that is a surprise, not a pass.

### Why it is inert on the two cases it was filed under

Probed the real gathered+filtered pool and ran `simpleRerank` with the pass ON and OFF. **Byte-identical winners both ways.** The sibling precondition is met and is not the blocker — `apple` finds 10 nutrition-bearing siblings, `unsweetened almond milk` finds 9. The tuning is:

- **`n-prod-01` (`apple`)** — sibling kcal `19, 23.9, 46.6, 46.6, 49, 52, 57, 61, 71, 398`, median `50.5`. Within `CONSENSUS_AGREE_TOLERANCE` (±15%) exactly **5 of 10** agree, and the guard is `agreeing.length * 2 <= values.length` — a strict majority. 5/10 fails by one record.
- **`n-mod-02` (`unsweetened almond milk`)** — median `19.4`, below `CONSENSUS_MIN_KCAL_MEDIAN = 50`, so `outliersOf()` returns empty before any comparison. That floor is deliberate and correct: on a near-zero-calorie food, ratios are noise.

So the brand gate was never what stopped this pass on generic queries. **The tuning is the binding constraint**, and it was calibrated for tight branded sibling groups (the "42% protein Quest bar" class). Generic produce pools are genuinely dispersed. Removing the gate without re-deriving the tuning changes nothing on the population it was filed against.

### The part that makes it actively unsafe

Swept all 160 golden nlp queries through gather + filter + `simpleRerank`, consensus ON vs OFF, no LLM calls. **The winner changes on 14 of 159 queries with a non-empty pool — and 13 of the 14 move FatSecret -> OpenFoodFacts.**

```
n-gen-01  chicken breast    fs_1641  -> off_5010525087528
n-gen-02  brown rice        fs_4513  -> off_0081652000549
n-prot-01 chicken thigh     fs_1695  -> off_0213580011124
n-svg-08  instant oatmeal   fs_4490  -> off_0688267185342
n-serv-02 oats              fs_39715 -> off_0299649508392
n-serv-04 white rice        fs_4501  -> off_0074265017325
n-tot-02  white rice        fs_4501  -> off_0074265017325
n-serv-39 mayonnaise        fs_34194 -> off_0060383995188
n-mq-36   oatmeal           fs_4481  -> off_0062600962836
...
```

A FatSecret record is typically a singleton inside a sibling group dominated by OFF rows, so it reads as the outlier against an OFF-majority median and takes the penalty. That is the wrong direction for a pipeline whose documented dominant defect is OFF rows winning when they should not. Note `n-tot-02` and `n-serv-04`: both would flip off `fs_4501`, the exact record PR #234 was built to reach.

(That sweep passes `targetBrand=undefined` for every query, whereas production detects a brand for some, so 14/159 is an **upper bound** on live winner changes. The direction is the finding, not the magnitude.)

### What was actually wrong with the approach

I should have run `scripts/eval/winner-diff.ts` before opening this. Its own header names this exact pass as one of four non-monotone consumers a ranking change must be diffed against, and the investigation playbook already demands a winner diff before any ranking change may be called safe. I hand-rolled a weaker version of it after the fact — and unlike `winner-diff`, mine had no Prisma write guard.

The 29,394-row population measurement that motivated this stands. What is refuted is that ungating the pass reaches it.

**Reviving this needs, in order:** (1) re-derive `CONSENSUS_AGREE_TOLERANCE` / `CONSENSUS_MIN_KCAL_MEDIAN` against real generic pools rather than branded ones; (2) make the pass source-aware so a lone cross-source record cannot be outvoted by same-source siblings; (3) gate on `winner-diff`, not on the golden set alone — the golden set only caught 2 of the 14 winner changes.

evawlve:
Closing rather than leaving open. The measurement is preserved in three places and none of them is this PR:

- `sync-docs/pipeline_hardening_plan.md` §Phase 2 item 4 (mobile repo) owns the full result — the brand gate was never the binding constraint, the tuning is; and where the ungated pass does act, 13 of 14 winner changes move FatSecret → OpenFoodFacts, i.e. the wrong direction.
- The branch `rerank/consensus-generic-queries` is untouched and can be reopened with `gh pr create --head rerank/consensus-generic-queries`.
- Reviving this needs, in order: re-derive `CONSENSUS_AGREE_TOLERANCE` and `CONSENSUS_MIN_KCAL_MEDIAN` against *generic* pools; make the pass source-aware so a lone cross-source record cannot be outvoted by same-source siblings; gate on `winner-diff`, not the golden set (it caught 2 of the 14).

An open PR titled DO NOT MERGE is a trap across a context reset — the title is the part that gets read. Closed is the safer encoding of 'parked'.

PR #255, opened 2026-08-06
fix(rerank): lane identity is source × retrieval mode, nested under source

evawlve:
Plan item **#18**. Owner: `sync-docs/pipeline_hardening_plan.md` §Phase 2 #18 (mobile repo).

## What it does

`buildRerankPool()` groups candidates by `source`, but the OFF lane is **two concatenated
already-sorted blocks** — keyword hits, then semantic hits — and the concatenation is never
re-sorted. So whenever the keyword block alone fills OFF's budget, the semantic block is
**structurally unreachable**: admitted, never scored. That is the module's own founding defect
(lane starvation) one level down, within-source and cross-retrieval-mode.

The repair is lane identity = **source × retrieval mode**, and the mode split is **NESTED**
under `source` rather than flat. Outer round-robin still runs over `source`, so every source
keeps exactly today's share (new **invariant 6**); the inner rotation alternates a source's
retrieval blocks. `semanticSimilarity` is already a first-class field on `UnifiedCandidate`.

Nesting is not cosmetic — the flat form (`lane = source + '#' + mode`) gave OFF two slots per
pass, which dropped FatSecret 3 → 2, and FatSecret rows are the ones carrying label servings.
That cost 57 of 108 cold winner changes moving fatsecret → openfoodfacts with **zero** moving
back. Nesting restores **invariants 3, 4 and 5 in full**, which also dissolves this item's
stated prerequisite: the FDC nutrition anchor cannot move, so the anchor-decoupling PR is not
needed for this change.

## Gate

Population: 253 cold seeds enriched for this defect (generic OFF names with ≥8 branded and ≥8
embedded siblings, cut from `OffFood`) plus 600 real-traffic queries (`--from-events`), one
snapshot shared by every arm, `--with-serving` on both replays and all noise floors.
**Noise floor 0 winner-diffs / 0 path-diffs on every tree.**

- Identity moves as designed: branded → generic **74:1** cold, **46:4** traffic.
- Re-gated 2026-08-05 as a clean 2×2 on `wd-snap-n1.json` (1,136 queries), four distinct tree
  hashes, `caller`/`helpers` drift hashes identical on all four. The control validates: the
  nested-only arm reproduces the recorded traffic figures on an independently re-cut
  population — GRAMS-CHANGED **7.5% exactly**, identity **42:4** against the recorded 46:4.
- Discriminator misattribution (gather's dedupe keeps the keyword copy and merges the semantic
  flag onto it) is **71 of 1,745 flagged OFF candidates (4.1%)**, costing 14 of 382 window
  slots nested. Invariant-6 violations: **0 of 241 rows**.

## The residual, and why it is now a policy bound rather than a gate artefact

The identity win and the serving loss are **the same event**: a generic OFF row is the better
answer to a generic query, and roughly half of them carry no `servingGrams`, so serving
resolution falls through to the 100 g-default family.

That residual has narrowed three times, and never for the reason this item originally claimed
(it was never serving coverage — every affected row has ≥3 in-band name siblings):

| | residual | mechanism removed |
|---|---|---|
| as built | 35 | — |
| + #250 (N1, name-group sibling anchor) | 28 | siblings, partially |
| + #252 (tight-name-group clamp relaxation) | 15 | N1's raise-only clamp |
| + #254 (bare-plural takes the tight test too) | **8** | N1's bare-plural suppression |

**All 8 remaining rows are dispersed name groups** (IQR ratio 1.69–4.33) — exactly the
population the clamp exists to refuse. The residual is bounded by measured policy now, and
cannot fall below 8 without changing that policy.

#254 was gated on both arms: **24 of 1,136 rows move on master, 28 on master+#18, ZERO winner
changes in both**, and it repaired **exactly the 7** of #18's 13 plural-blocked rows predicted
from the tightness table before the gate ran.

## Scope

Two files: `src/lib/mapping/rerank-pool.ts` and its test. Nothing under `data/` or `prisma/`,
so no hand-`scp` and no migration. Merges clean onto master — the `_lane_identity_probe.ts`
add/add conflict recorded in the plan was resolved on this branch in `5e33668` by taking
master's copy as canonical.


evawlve:
## Correction to this PR's description, and the post-deploy result

**Correction — I quoted two different arms as if they were one.** The description says
"branded → generic **74:1** cold, **46:4** traffic". Those are *different trees*: 74:1 cold and
48:7 traffic are the **FLAT** variant, which was gated and thrown away; 46:4 (re-cut 42:4) is the
**NESTED** variant, which is what merged. **The nested form's COLD identity figure does not exist
in any doc.** Same class of error as the one the plan's §Phase 2 #18 entry already records
against itself — pricing a cost from one tree against a benefit from another.

## Deployed and gated

Deployed with #254 in one build: `5a5S6vnOghRdfp2zptZ3R` → `EHel-JzvULhCLmAinwoLM`. Three
box-served cold runs with a `systemctl --user restart recipe-api` between each; `buildId`
identical on both sides of every run. **Stable set 12 → 11; noise floor 2/348 on this
sequence.**

**The pre-registered prediction landed.** Before deploying I predicted `n-dens-05` would move,
because #18's mechanism is exactly its documented cause (the OFF lane spends its slots on
branded rows before reaching the eight brandless `Rolled Oats` records that score identically).
It moved, to passing.

| case | query | before | after |
|---|---|---|---|
| `n-dens-05` | generic `rolled oats` | branded Bob's Red Mill, 96 g ✗ | generic `Rolled oats`, 86.4 g ✓ |
| `n-svs-05` | generic `coca cola` can | `Coca Cola` 300 g ✗ | `Coca cola` 355 g ✓ |
| `n-mq-41` | **chain** `noodles and company pad thai` | `fs_81276` Pad Thai (Small) [**Noodles & Company**] 260 g ✓ | `off_4099100064773` Pad Thai Noodles [**NULL brand**] 127.5 g ✗ |

## The finding, which is bigger than this PR

**#18 helps when the query is generic and hurts when the query names a brand.** Both repairs are
generic queries that were resolving to branded records — the defect #18 was built for. The one
regression is a query that *names a restaurant chain*, where the old winner was the correct
branded chain dish and the new winner is a brandless packaged noodle. Settled by reading
`MappingEventLog`, not the `detail` string.

That is consistent with what this repo already documents — chains fail identity **31.5%** vs
branded CPG **under 9%** — and it means the "branded → generic" direction was counted, never
adjudicated for correctness on brand-bearing queries.

Two things the frozen-pool gate structurally could not have shown, both now demonstrated:
1. `winner-diff` aborts at winner selection (its own limitation **B**), so a winner change that
   is fine on identity but wrong on the *assertion* is invisible to it.
2. `n-mq-41` is one of the exact **7 rows #254 was credited with repairing** under #18. The
   repair was real but was *tier-family membership only* — the plan's own "read 'repairs'
   narrowly: only 2 of the 7 restore master's billed grams exactly". This is one of the other 5,
   and it now has a live gating consequence.

**Not reverted**, because the net golden-set movement is favourable (two repairs against one
regression) and the residual was the priced trade. But the framing Diego approved understated
the cost, and the brand-bearing-query case is now the top open question against this change.

